PR #20, opened 2019-08-20
CPDP-503: retrieve additional data for registration

co-category:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/connectors/CGTPropertyDisposalsConnector.scala
```diff
@@ -30,33 +30,42 @@ import scala.concurrent.{ExecutionContext, Future}
...
   val subscribeUrl: String = s"$baseUrl/subscribe"
 
-  def getBusinessPartnerRecord(nino: NINO)(implicit hc: HeaderCarrier): EitherT[Future, Error, HttpResponse] =
+  def getBusinessPartnerRecord(nino: NINO, name: Name, dob: DateOfBirth)(
+    implicit hc: HeaderCarrier
+  ): EitherT[Future, Error, HttpResponse] =
     EitherT[Future, Error, HttpResponse](
-      http.get(bprUrl(nino))
+      http
+        .get(bprUrl(nino))
```
AndyHWChung: isn't this a POST with a body now?

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/DateOfBirth.scala
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.cgtpropertydisposalsfrontend.models
+
+import org.joda.time.LocalDate
```
AndyHWChung: can we use java `LocalDate` please? joda is deprecated

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/connectors/CGTPropertyDisposalsConnector.scala
co-category: Yes! Good spot. The backend endpoint changed late yesterday.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/DateOfBirth.scala
co-category: changed - erroneous import in Intellij

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/BusinessPartnerRecord.scala
```diff
@@ -19,11 +19,11 @@ package uk.gov.hmrc.cgtpropertydisposalsfrontend.models
 import play.api.libs.json.{Format, Json}
 
 final case class BusinessPartnerRecord(
-    forename: String,
-    surname: String,
-    emailAddress: Option[String],
-    address: Address,
-    sapNumber: String
+  forename: String,
```
AndyHWChung: you might need to get rid of `forename` and `surname` ub this case class? The BE won't be sending it in the response since you took it out

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/DateOfBirth.scala
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.cgtpropertydisposalsfrontend.models
+
+import java.time.LocalDate
+
+import play.api.libs.json.{Json, OFormat}
+
+final case class DateOfBirth(value: LocalDate)
```
AndyHWChung: this isn't anything major, but you could do an optimisation here by extending `AnyVal` - this will make it so that at certain times instances of this case class aren't wrapped in a class. See here:

https://docs.scala-lang.org/overviews/core/value-classes.html

The same thing applies for things like implicit classes as well

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/BusinessPartnerRecord.scala
co-category: Yes. Will modify the `SubscriptionDetails.fromBusinessPartnerRecord` accordingly to make the auth record the source of truth for the name of the individual

co-category: refactored

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/DateOfBirth.scala
co-category: converted to a value class

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/actions/AuthenticatedAction.scala
```diff
@@ -71,48 +78,82 @@ class AuthenticatedAction @Inject() (
...
-        handleInsufficientConfidenceLevel(request, errorHandler.errorResult()(request)).map(Left(_))
-
+  private def makeAuthenticatedRequestWithRetrieval[A](
+    nino: String,
+    name: RetrievalName,
+    dateOfBirth: LocalDate,
+    request: MessagesRequest[A]
+  ): Future[Either[Result, AuthenticatedRequest[A]]] =
+    name match {
+      case RetrievalName(Some(forename), Some(surname)) =>
```
AndyHWChung: this is where the auth team said that the full name goes in the `name` field. So we'd have to do something like:
```scala
    case RetrievalName(Some(fullName), _) =>
      fullName.split(' ').toList match {
        case firstName :: (_ :+ lastName) => // stuff
        case _                            => // error
      }
```

AndyHWChung: not a great case class to work with....

co-category: I also don't like this:

```scala
 private def makeAuthenticatedRequestWithItmpRetrieval[A](
    nino: String,
    name: ItmpName,
    dateOfBirth: LocalDate,
    request: MessagesRequest[A]
  ): Future[Either[Result, AuthenticatedRequest[A]]] =
    name match {
      case ItmpName(Some(givenName), _, Some(familyName)) =>
        Right(AuthenticatedRequest(NINO(nino), Name(givenName, familyName), DateOfBirth(dateOfBirth), request))
      case other =>
        logger.warn(s"Failed to retrieve the complete name for this user: $other")
        Left(errorHandler.errorResult()(request))
    }

  private def makeAuthenticatedRequestWithRetrieval[A](
    nino: String,
    name: RetrievalName,
    dateOfBirth: LocalDate,
    request: MessagesRequest[A]
  ): Future[Either[Result, AuthenticatedRequest[A]]] =
    name match {
      case RetrievalName(Some(forename), Some(surname)) =>
        Right(AuthenticatedRequest(NINO(nino), Name(forename, surname), DateOfBirth(dateOfBirth), request))
      case other =>
        logger.warn(s"Failed to retrieve the complete name for this user: $other")
        Left(errorHandler.errorResult()(request))
    }
```

I want to add a type parameter the method so that a sing;e method does this but the field names are different for each type.

co-category: > this is where the auth team said that the full name goes in the `name` field. So we'd have to do something like:
> 
> ```scala
>     case RetrievalName(Some(fullName), _) =>
>       fullName.split(' ').toList match {
>         case firstName :: (_ :+ lastName) => // stuff
>         case _                            => // error
>       }
> ```

Yeah I realised I hadn't done that bit as soon as I popped out for lunch!

AndyHWChung: the other option is to have a type signature which looks something like this:
```
def makeAuthenticatedRequestWithRetrieval[A](
    nino: String,
    firstName: String,
    lastName: String,
    dateOfBirth: LocalDate,
    request: MessagesRequest[A]
  ): Future[Either[Result, AuthenticatedRequest[A]]] =
```
and move the getting of the first and last name outside of the function.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/actions/AuthenticatedAction.scala
```diff
@@ -71,48 +78,82 @@ class AuthenticatedAction @Inject() (
...
     }
+
+  @SuppressWarnings(Array("org.wartremover.warts.Product", "org.wartremover.warts.Serializable"))
+  override protected def refine[A](request: MessagesRequest[A]): Future[Either[Result, AuthenticatedRequest[A]]] = {
+    implicit val hc: HeaderCarrier =
+      HeaderCarrierConverter.fromHeadersAndSession(request.headers, Some(request.session))
+
+    authorisedFunctions
+      .authorised(ConfidenceLevel.L200)
+      .retrieve(v2.Retrievals.nino and v2.Retrievals.itmpName and v2.Retrievals.name and Retrievals.itmpDateOfBirth) {
```
AndyHWChung: could we not use `v2.Retrievals.itmpDateOfBirth` as well?

co-category: yes, should have been v2.Retrievals.itmpDateOfBirth

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/actions/AuthenticatedAction.scala
co-category: refactored this completely

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/actions/AuthenticatedAction.scala
```diff
@@ -71,48 +76,97 @@ class AuthenticatedAction @Inject() (
...
+      case ItmpName(Some(givenName), _, Some(familyName)) => Some(Name(givenName, familyName))
+      case _                                              => None
+    }
 
-      case _: InsufficientConfidenceLevel =>
-        handleInsufficientConfidenceLevel(request, errorHandler.errorResult()(request)).map(Left(_))
+  private def nonItpmNameToName(retrievalName: RetrievalName): Option[Name] =
+    retrievalName match {
+      case RetrievalName(Some(name), _) => {
+        val Array(forename, surname) = name.split(' ')
```
AndyHWChung: we also need to handle the case where there might be middle names inside this name. We need to look for the first and last element after the split.

co-category: refactored to handle arbitrary names - went for more verbose logic to make the logic apparent

AndyHWChung:
LGTM

PR #27, opened 2019-08-28
 cpdp-502-523-524: check affinity group; ask if individual has nino if cl < 200

AndyHWChung:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/IvController.scala
```diff
@@ -18,31 +18,44 @@ package uk.gov.hmrc.cgtpropertydisposalsfrontend.controllers
...
-
-      case None =>
-        logger.warn("Could not get ivContinueUrl from session")
-        InternalServerError(errorHandler.internalServerErrorTemplate)
+                            )(implicit ec: ExecutionContext) extends FrontendController(cc)
+  with WithAuthAndSessionDataAction
+  with Logging
+  with SessionUpdates {
+
+  def ivSuccess(): Action[AnyContent] = authenticatedActionWithSessionData.async { implicit request =>
```
AndyHWChung: I made it so that we always go to the start endpoint after successful IV. Someone who had to go through the IV journey would not have been going through anywhere other than the start endpoint since there would be no session data to say that they should be somewhere further on in the journey and all of our endpoints should be protected with logic to make sure the session data says they're in the right place.

PR #29, opened 2019-09-02
CPDP-514: add tax enrolment integration

co-category:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/services/TaxEnrolmentService.scala
```diff
@@ -0,0 +1,69 @@
...
+  sealed trait TaxEnrolmentResponse
+
+  object TaxEnrolmentResponse {
+    case object TaxEnrolmentCreated extends TaxEnrolmentResponse
+  }
+
+}
+
+@Singleton
+class TaxEnrolmentServiceImpl @Inject()(connector: TaxEnrolmentConnector) extends TaxEnrolmentService {
```
AndyHWChung: do you a spec for this or am I being blind?

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/TaxEnrolmentSupport.scala
```diff
@@ -0,0 +1,41 @@
...
+
+import cats.data.EitherT
+import org.scalamock.scalatest.MockFactory
+import uk.gov.hmrc.cgtpropertydisposalsfrontend.connectors.TaxEnrolmentConnector
+import uk.gov.hmrc.cgtpropertydisposalsfrontend.models.{EnrolmentRequest, Error}
+import uk.gov.hmrc.http.{HeaderCarrier, HttpResponse}
+
+import scala.concurrent.Future
+
+trait TaxEnrolmentSupport {
```
AndyHWChung: I'm not sure any other spec other for anything other than `TaxEnrolmentService` will ever use this stuff since that is the only thing in charge of enroling

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/services/TaxEnrolmentService.scala
```diff
@@ -0,0 +1,69 @@
...
+
+  override def allocateEnrolmentToGroup(subscriptionResponse: SubscriptionResponse)(
+    implicit hc: HeaderCarrier
+  ): EitherT[Future, Error, TaxEnrolmentResponse] = {
+    val emptyEnrolmentRequest = EnrolmentRequest(List.empty, List.empty)
```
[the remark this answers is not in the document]
co-category: Yes, but waiting for the DES specs to come through and then this will be updated. Unless you know for sure what they are.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/services/TaxEnrolmentService.scala
co-category: Yeah, search for TaxEnrolmentConnectorImplSpec

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/TaxEnrolmentSupport.scala
co-category: OK, will rip this out.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/services/TaxEnrolmentService.scala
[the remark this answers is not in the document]
AndyHWChung: we should be pretty sure what the identifiers and verifiers are. The commit above in the enrolment config is what was configured for our service. That's derived from the response to subscription in the EPID

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/services/TaxEnrolmentService.scala
AndyHWChung: isn't that the spec for the connector rather than the service?

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/services/TaxEnrolmentService.scala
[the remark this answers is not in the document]
co-category: OK, so I will take the postcode and countrycode from the SubscriptionDetails.Address field.

co-category:
It depends on your retry logic. What do you intend to do?

AndyHWChung:
I had an idea to make use of the akka actor framework to have an actor managing all the retries. Actors are nice in this scenario since they can handle scheduling and they can handle mutable state nicely as well.

co-category:
Yes you can use the circuitbreaker feature to model this. In my view there is no persistence then, you will retry based on some schedule.

AndyHWChung:
persistence will be useful so that we can keep on retrying if we restart the service. Otherwise, the in memory retries will be lost after the service shuts down

co-category:
What will be your retry schedule? Days, weeks, months?

co-category:
Will transfer this implementation to the back-end.

PR #30, opened 2019-09-02
cpdp-521: allowed trusts to subscribe

AndyHWChung:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/StartController.scala
```diff
@@ -95,19 +103,49 @@ class StartController @Inject()(
...
+      for{
+        bprWithNameAndEmail <- bprService.getBusinessPartnerRecord(Left(trust))
+          .subflatMap[Error,(BusinessPartnerRecord,TrustName,Email)]{ bpr =>
+          (bpr.organisationName, bpr.emailAddress) match {
+            case (None, None) =>
+              Left(Error("Could not find trust name or email address from business partner record"))
+            case (Some(_), None) =>
+              Left(Error("Could not find trust name from business partner record"))
+            case (None, Some(_)) =>
+              Left(Error("Could not find email address from business partner record"))
```
AndyHWChung: we should probably handle this situation like we do for individuals: we should try and get the gg email from auth and if that still doesn't exist then send them down the enter email journey which includes email verification. That can come in a later piece of work though - too much in this PR otherwise

AndyHWChung: I wrote CPDP-571 for this

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/connectors/CGTPropertyDisposalsConnector.scala
```diff
@@ -30,7 +34,7 @@ import scala.concurrent.{ExecutionContext, Future}
 @ImplementedBy(classOf[CGTPropertyDisposalsConnectorImpl])
 trait CGTPropertyDisposalsConnector {
 
-  def getBusinessPartnerRecord(nino: NINO, name: Name, dob: DateOfBirth)(
+  def getBusinessPartnerRecord(entity: Either[Trust,Individual])(
     implicit hc: HeaderCarrier
```
co-category: Wouldn't this have been better as a sub-type of a sealed trait? That would allow us to easily extend the code for other types of users e.g. agents.

etq:q
```diff
@@ -1,251 +0,0 @@
-[33mcommit a3aca1357b6126d776d7706fdfbc9196084534a6[m[33m ([m[1;36mHEAD -> [m[1;32mretrieve-data-in-one-place[m[33m)[m
```
co-category: what is the file doing here?

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/connectors/CGTPropertyDisposalsConnector.scala
AndyHWChung: individuals and trusts should be the only two types of entities we get a bpr using register with id for. We wouldn't get a business partner for an agent since we don't allow them to go through a subscription journey - the client needs to go through the journey themselves

etq:q
AndyHWChung: I deleted it - not sure how it got in there. Looking at the commit history looks like I somehow generated it and commited it and push it up accidentally. Looking at the file name, looks like I was trying to enter a vim command somewhere but I wasn't typing it in the right place

PR #32, opened 2019-09-06
cpdp-521: register you trust screen can now be reached

AndyHWChung:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/SubscriptionStatus.scala
```diff
@@ -17,13 +17,13 @@
...
 import julienrf.json.derived
-import play.api.libs.json.{Format, JsDefined, JsNull, JsObject, JsResult, JsSuccess, JsUndefined, JsValue, Json, OFormat, OWrites, Reads, Writes}
+import play.api.libs.json.OFormat
 
 sealed trait SubscriptionStatus
 
 object SubscriptionStatus {
 
-  final case object OrganisationUnregisteredTrust extends SubscriptionStatus
+  final case object NonTrustOrganisation extends SubscriptionStatus
```
co-category: I'm getting confused again...is this an organisation which is really a trust but has never enrolled as one with HMRC? That is, a pre-enrollment era trust? If so I think the "OrganisationUnregisteredTrust" is a more reflective name or something like UnenrolledTrust or UnregisteredTrust. NonTrustOrganisation seems to imply that it is an organisation coming in but is not a trust (enrolled or not enrolled).

AndyHWChung: We can't distinguish between legacy trusts and organisations which aren't trusts. The class represents any organisation which doesn't have a trust enrolment - so it could be either one. So perhaps the name before was a bit more accurate

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/SubscriptionStatus.scala
```diff
@@ -17,13 +17,13 @@
...
 
   final case class IndividualInsufficientConfidenceLevel(hasNino: Option[Boolean], hasSautr: Option[HasSAUTR]) extends SubscriptionStatus
```
co-category: Can we call this IndividualWithInsufficientConfidenceLevel ?

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/actions/AuthenticatedActionWithRetrievedData.scala
```diff
@@ -154,7 +155,7 @@ class AuthenticatedActionWithRetrievedData @Inject()(
   // work out if it is an organisation or not
     enrolments.getEnrolment("HMRC-TERS-ORG") match {
       case None =>
-        Left(SeeOther(routes.RegisterTrustController.registerYourTrust().url))
+        Right(AuthenticatedRequestWithRetrievedData(UserType.NonTrustOrgansation, request))
```
co-category: Shouldn't UserType.NonTrustOrganisation change to UserType.OrganisationUnregisteredTrust?

AndyHWChung: yup, renamed it

PR #33, opened 2019-09-06
CPDP-584: added manual address page, links, routes and filter field

nubz:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/AddressLookupResult.scala
```diff
@@ -20,6 +20,7 @@ import play.api.libs.json.{Format, Json}
 
 final case class AddressLookupResult(
   postcode: Postcode,
+  filter: Option[String],
   addresses: List[Address]
```
co-category: why does filter need to be in the AddressLookupResult? The result should just contain the address or nothing.

co-category:
did you run this locally? also could you add the address-lookup dependency to the service manager profile (you will prolly also need the address-lookup-stub to test it against the locally running service)

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/AddressController.scala
```diff
@@ -44,30 +46,71 @@ class AddressController @Inject()(
...
-      Postcode.form
+      AddressLookupRequest.form
         .bindFromRequest()
         .fold(
-          formWithErrors => BadRequest(enterPostcodePage(formWithErrors, routes.SubscriptionController.checkYourDetails)), { postcode =>
-            if (request.sessionData.addressLookupResult.map(_.postcode).contains(postcode)) {
+          formWithErrors => BadRequest(enterPostcodePage(formWithErrors, routes.SubscriptionController.checkYourDetails())),
+          { case AddressLookupRequest(postcode, filter) =>
+            // TODO: need to search session data with filter also
+            if (request.sessionData.addressLookupResult.map(_.postcode).contains(postcode)
```
co-category: remove this TODO as you are doing the lookup

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/AddressLookupResult.scala
AndyHWChung: it might be being used to prepopulate the form is a user goes back, and also on the form submit endpoint to check if someone's already we've got a search result for the parameters the user's entered. In that case we can skip a call to address lookup and just use the cached address lookup result.

co-category:
do you have any tests where the filter is other than None? I don't think I can see any

AndyHWChung:
locally we're currently just using our own stub to stub out the address lookup service - I can't remember if we did a proper investigation on if the address lookup stub is easy to use or whether we need to have special test data in mind

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/AddressController.scala
nubz: done

nubz:
> do you have any tests where the filter is other than None? I don't think I can see any

there is a test in the AddressConnectorSpec but now I have added 2 more into the AddressControllerSpec also to handle submitted filter values including handling empty string

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/AddressLookupResult.scala
nubz: It's useful to hold in session so that the exact search that was used to produce the results can be replayed.

nubz:
> did you run this locally? also could you add the address-lookup dependency to the service manager profile (you will prolly also need the address-lookup-stub to test it against the locally running service)

Works locally fine - now updated conf to use ADDRESS_LOOKUP_STUB which works fine too :)

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/AddressController.scala
```diff
@@ -44,30 +46,70 @@ class AddressController @Inject()(
                                    addressLookupService: AddressLookupService,
                                    sessionStore: SessionStore,
                                    enterPostcodePage: views.html.subscription.enter_postcode,
-                                   selectAddressPage: views.html.subscription.select_address
+                                   selectAddressPage: views.html.subscription.select_address,
+                                   enterAddressPage: views.html.subscription.enter_address,
+                                   addressDisplay: views.html.components.address_display
 )(implicit viewConfig: ViewConfig, ec: ExecutionContext)
```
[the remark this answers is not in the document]
nubz: Done

PR #42, opened 2019-09-12
cpdp-589: bpr retrieved using sautr when available for user with cl<200

AndyHWChung:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/StartController.scala
```diff
@@ -118,21 +118,30 @@ class StartController @Inject()(
   }
 
 
-  private def handleInsufficientConfidenceLevel(maybeNino: Option[NINO], name: Name, maybeEmail: Option[Email])(
+  private def handleInsufficientConfidenceLevel(maybeNino: Option[NINO],
+                                                maybeSautr: Option[SAUTR],
+                                                name: Name,
+                                                maybeEmail: Option[Email])(
     implicit request: RequestWithSessionDataAndRetrievedData[AnyContent]
```
co-category: formatting

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/UserType.scala
```diff
@@ -33,6 +33,7 @@ object UserType {
 
   final case class IndividualWithInsufficientConfidenceLevel(
                                                               nino: Option[NINO],
+                                                              sautr: Option[SAUTR],
                                                               name: Name,
                                                               email: Option[Email]
                                                             ) extends UserType
```
co-category: formatting

AndyHWChung: same here

co-category: The scalafmt config should force this class to be formatted to something like:

```scala
@Singleton
class BusinessPartnerRecordConnectorImpl @Inject()(
  http: HttpClient,
  val config: ServicesConfig
)(implicit ec: ExecutionContext)
```

ie the class variables are two spaces indented from the very left.

AndyHWChung: oh - not sure why scalafmt is not doing it automatically. Unless we're missing a setting or something. I'll change it

PR #45, opened 2019-09-16
Cpdp 508 capture name

AndyHWChung:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/AddressController.scala
```diff
@@ -172,7 +171,7 @@ class AddressController @Inject()(
 
   private def storeAddress(address: Address)(implicit request: RequestWithSubscriptionReady[_]): Future[Result] =
     updateSession(sessionStore, request)(
-      _.copy(subscriptionStatus = Some(addressLens.set(request.subscriptionReady)(address)))
+      _.copy(journeyStatus = Some(addressLens.set(request.subscriptionReady)(address)))
     ).map(
```
co-category: This is confusing - on the one hand an address is being set but it is being assigned to an identifier which has name subscriptionStatus or now journeyStatus - can this be improved to make the intention clear?

AndyHWChung: the subscriptionReady is being updated with a new address with the `Lens` called `addressLens` - if it's too confusing I can take the lens stuff out to make it clearer what's going on. I just included it to avoid having to do stuff like `A.copy(a = A.a.copy(...))`

AndyHWChung: do you reckon this is too ugly?
```scala
      _.copy(journeyStatus = Some(request.subscriptionReady.copy(
        subscriptionDetails = request.subscriptionReady.subscriptionDetails.copy(
          address = address
        ))))
```

co-category: no, let's keep it in. Lens are fine. I looked at monocle and I don't think we will get the syntax any terser than what you have here already.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/DefaultRedirects.scala
```diff
@@ -16,25 +16,40 @@
...
 
+  private def subscriptionDefaults(subscriptionStatus: SubscriptionStatus): Call = {
+    import SubscriptionStatus._
+    subscriptionStatus match {
+    case _: SubscriptionMissingData  => routes.StartController.start()
+    case _: SubscriptionReady        => routes.SubscriptionController.checkYourDetails()
+    case _: SubscriptionComplete     => routes.SubscriptionController.subscribed()
+    case _: IndividualWithInsufficientConfidenceLevel => routes.InsufficientConfidenceLevelController.doYouHaveNINO()
+    case OrganisationUnregisteredTrust  => routes.RegisterTrustController.registerYourTrust()
+    }
```
co-category: Let's rename OrganisationUnregisteredTrust to UnregisteredTrust - that fact it is a trust implies it is an organisation

AndyHWChung: done

AndyHWChung:
I forgot to mention on this PR: I started extending sealed traits with `Product with Serializable` for better type inference: https://nrinaudo.github.io/scala-best-practices/adts/product_with_serializable.html

AndyHWChung:
if we don't have it sometimes the compiler will infer a `Product` type and then wart remover complains and doesn't allow you to compile the code

co-category:
Cool - that should help us to avoid wart remover warnings as well.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/JourneyStatus.scala
```diff
@@ -0,0 +1,77 @@
...
+
+sealed trait JourneyStatus extends Product with Serializable
+
+
+object JourneyStatus {
+
+  ////
+  // Subscription Journey
+  ////
+  sealed trait SubscriptionStatus extends JourneyStatus with Product with Serializable
```
co-category: these ugly comments are unnecessary - code documents itself here.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/JourneyStatus.scala
```diff
@@ -0,0 +1,77 @@
...
+                                           subscriptionResponse: SubscriptionResponse)
+      extends SubscriptionStatus
+
+  }
+
+
+  ////
+  // Subscription Journey
+  ////
+  sealed trait RegistrationStatus extends JourneyStatus with Product with Serializable
```
co-category: same here as well

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/JourneyStatus.scala
AndyHWChung: yeah, they're a bit ugly - I'll take them out

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/JourneyStatus.scala
AndyHWChung: taken out now

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/JourneyStatus.scala
```diff
@@ -0,0 +1,77 @@
...
+
+    final case object IndividualWantsToRegisterTrust extends RegistrationStatus
+
+    final case class IndividualSupplyingInformation(name: Option[Name]) extends RegistrationStatus
+
+  }
+
+
+  @SuppressWarnings(Array("org.wartremover.warts.PublicInference"))
+  implicit val format: OFormat[JourneyStatus] = derived.oformat[JourneyStatus]
```
co-category: this has a type ascription so why the suppression of this wart error?

AndyHWChung: I'm not sure why this happens - it complains about it if you don't suppress the warning here. It will have something to do with how the play-json-codec library is doing its magic with `derived`

co-category: but it doesn't happen with the other codecs which are defined in this service so something peculiar happening here...

PR #51, opened 2019-09-27
Cpdp 600 autocomplete patch

nubz:


public/javascript/cgtpd.js
```diff
@@ -1,8 +1,27 @@
...
+    url: '/cgt-property-disposals/assets/location-autocomplete-graph.json',
+    defaultValue: ''
   });
+  // patch to ensure clearing the pseudo input also clears the select element
+  // this is an active issue on accessible-autocomplete
+  // https://github.com/alphagov/accessible-autocomplete/issues/260
+  // when it is resolved we can remove this code
+  var wrapper = document.querySelector('.country-code-wrapper');
+  function resetSelectIfEmpty (e) {
+    if (e.target.id === 'countryCode' && e.target.value === '') {
```
AndyHWChung: does `e.target.value` correspond to what the user's entered into the autocomplete box? If it is, is it worth trimming it to catch out the cases when they just press space?

nubz: done

nubz: Now updated to handle non matching strings - they all set the select to empty string instead of preserving previous selections - also removed UK/GB from graph

AndyHWChung:
LGTM

PR #52, opened 2019-09-27
cpdp-509: address now asked for in registration journey

AndyHWChung:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/JourneyStatus.scala
```diff
@@ -59,7 +60,9 @@ object JourneyStatus {
 
     final case object IndividualWantsToRegisterTrust extends RegistrationStatus
 
-    final case class IndividualSupplyingInformation(name: Option[Name]) extends RegistrationStatus
+    final case class IndividualSupplyingInformation(name: Option[Name], address: Option[Address]) extends RegistrationStatus
```
co-category: Would you mind changing this to IndividualWithCleanSkin? I know it sounds terrible but that is the common term that HMRC use to refer to such an individual.

PR #53, opened 2019-10-03
cpdp-512-647: cya registration;handle no email for registration

AndyHWChung:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/InsufficientConfidenceLevelController.scala
```diff
@@ -73,7 +72,7 @@ class InsufficientConfidenceLevelController @Inject()(
   )(implicit request: RequestWithSessionData[_]): Future[Result] =
     request.sessionData.flatMap(_.journeyStatus) match {
       case Some(i: IndividualWithInsufficientConfidenceLevel) => f(i)
```
co-category: As discussed, `IndividualWithInsufficientConfidenceLevel` seems to represent a user whereas what we are attempting to represent is an actual type of user journey. As such, this should be renamed to reflect that. A possible name to represent this part of the journey could be "DetermineAffinityGroup"

AndyHWChung: I renamed it to `TryingToGetIndividualsFootprint` - someone has this journey status when they have CL<200 and we can't find a NINO or SAUTR. We take them through a journey where we ask if they have a NINO or SAUTR and subsequently if they are trying to register an individual or trust

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/email/RegistrationChangeEmailControllerSpec.scala
```diff
@@ -0,0 +1,136 @@
...
+import play.api.test.FakeRequest
+import uk.gov.hmrc.cgtpropertydisposalsfrontend.controllers
+import uk.gov.hmrc.cgtpropertydisposalsfrontend.controllers.RedirectToStartBehaviour
+import uk.gov.hmrc.cgtpropertydisposalsfrontend.models.JourneyStatus.RegistrationStatus.RegistrationReady
+import uk.gov.hmrc.cgtpropertydisposalsfrontend.models.{Email, sample}
+
+import scala.concurrent.Future
+
+class RegistrationChangeEmailControllerSpec
+  extends EmailControllerSpec[RegistrationReady,RegistrationReady]
```
co-category: Do you think it would be worth splitting the registration and subscription controllers into separate packages?

AndyHWChung: yeah, that would make it cleaner I think. Then we could probably get rid of `Registration` and `Subscription` from the class names - though that'd make it ambiguous if we every used those names in isolation without the correct context

PR #64, opened 2019-10-14
CPDP-547: check user's enrolment/subscription status

co-category:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/actions/AuthenticatedActionWithRetrievedData.scala
```diff
@@ -60,87 +63,127 @@ class AuthenticatedActionWithRetrievedData @Inject()(
...
+      case Some(cgtEnrolment) =>
+        cgtEnrolment.getIdentifier("CGTPDRef") match {
+          case Some(cgtReference) => Future.successful(Right(Some(CgtReference(cgtReference.value))))
+          case None               => Future.successful(Left(errorHandler.errorResult()(request)))
+        }
+      case None =>
+        subscriptionService.hasSubscription().value.map {
+          case Left(_) => Left(errorHandler.errorResult()(request))
+          case Right(maybeCgtReference) =>
+            maybeCgtReference match {
```
AndyHWChung: do you need to pattern match on `maybeCgtReference`? Looks like you can probably just do a `leftMap` on the `hasSubscription` `EitherT` and then `.value` that

conf/app.routes
```diff
@@ -110,6 +110,6 @@ GET  /register-your-trust                 uk.gov.hmrc.cgtpropertydisposalsfronte
 
 
 ### Return journey
-GET  /return/start                        uk.gov.hmrc.cgtpropertydisposalsfrontend.controllers.ReturnsController.start()
+GET  /return/start                        uk.gov.hmrc.cgtpropertydisposalsfrontend.controllers.HomeController.start()
```
AndyHWChung: to avoid confusion we shoujld probably get rid of the references to returns in the comment and the url

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/StartController.scala
```diff
@@ -45,28 +44,30 @@ import scala.concurrent.{ExecutionContext, Future}
...
+    with WithAuthAndSessionDataAction
+    with Logging
     with SessionUpdates
     with IvBehaviour {
 
   def start(): Action[AnyContent] = authenticatedActionWithRetrievedDataAndSessionData.async { implicit request =>
-
     (request.authenticatedRequest.userType, request.sessionData.flatMap(_.journeyStatus)) match {
 
+      case (UserType.Subscribed(cgtReference, gGCredId), None) =>
```
AndyHWChung: would we not need a case for `(_, _: JourneyStatus.Subscribed)` as well?

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/SubscriptionController.scala
```diff
@@ -65,7 +63,8 @@ class SubscriptionController @Inject()(
         _ <- EitherT(
               updateSession(sessionStore, request)(
                 _.copy(journeyStatus = Some(SubscriptionComplete(details, subscriptionResponse)))
```
AndyHWChung: would this not be `JourneyStatus.Subscribed`? We don't really need `SubscriptionComplete` if we have that other journey status

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/JourneyStatus.scala
```diff
@@ -54,6 +54,12 @@ object JourneyStatus {
       subscriptionResponse: SubscriptionResponse
     ) extends SubscriptionStatus
 
+    // subscription has been submitted to ETMP
```
AndyHWChung: could we move this outside of the `SubscriptionStatus` sealed trait and have it under `JourneyStatus` or some other trait? When we come to displaying the homepage it'd be nice to use this `JourneyStatus` as the state to look for before displaying it - `SubscriptionStatus` ties it to the subscription journey but we should be able to get to this state without going through the subscription journey every time

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/HomeController.scala
```diff
@@ -23,14 +23,12 @@ import uk.gov.hmrc.cgtpropertydisposalsfrontend.views
...
 
 @Singleton
-class ReturnsController @Inject()(
+class HomeController @Inject()(
   cc: MessagesControllerComponents,
   startPage: views.html.returns.start
 )(implicit viewConfig: ViewConfig)
     extends FrontendController(cc) {
-
   def start(): Action[AnyContent] = Action { implicit request =>
```
AndyHWChung: we should only show this page if we're in the `JourneyStatus.Subscribed` state - we should go to the start endpoint otherwise. There's a method called `redirectToStartWhenInvalidJourney` in the unit tests to facilitate testing this.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/actions/AuthenticatedActionWithRetrievedData.scala
AndyHWChung: also you could just do `fold` on the `EitherT` rather than doing `.value` and having `case` statements

AndyHWChung: oh wait, this is equivalent to:
```scalasubscriptionService.hasSubscription().leftMap{
          e =>
          logger.warn("Could not check subscription status", e)
          errorHandler.errorResult()(request)
        }
```
(with an extra logging message)

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/JourneyStatus.scala
co-category: :thinking: super hard to know what to call it - any ideas? :confounded:

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/HomeController.scala
co-category:
```scala
  sealed trait SubscriptionComplete extends JourneyStatus with Product with Serializable

  object SubscriptionComplete {
    // subscription has been submitted to ETMP
    final case class Subscribed(
      subscriptionDetails: SubscriptionDetails,
      subscriptionResponse: SubscriptionResponse
    ) extends SubscriptionStatus
  }
```

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/JourneyStatus.scala
co-category:
```scala
  sealed trait SubscriptionComplete extends JourneyStatus with Product with Serializable

  object SubscriptionComplete {
    // subscription has been submitted to ETMP
    final case class Subscribed(
      subscriptionDetails: SubscriptionDetails,
      subscriptionResponse: SubscriptionResponse
    ) extends SubscriptionStatus
  }
```

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/actions/AuthenticatedActionWithRetrievedData.scala
co-category: I prefer this:

``scala
      case None =>
        subscriptionService
          .hasSubscription()
          .leftMap(_ => errorHandler.errorResult()(request))
          .value
```

File: app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/HomeController.scala (hunk 6 above)
Comment (co-category) [reply app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/HomeController.scala]: OK, I will create a trait `WithSubscriptionCompleteAction` and mix that in to the controller.

File: app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/HomeController.scala [hunk 7]
```diff
@@ -0,0 +1,60 @@
...
+  def start(): Action[AnyContent] = authenticatedActionWithRetrievedDataAndSessionData.async {
+    implicit request: RequestWithSessionDataAndRetrievedData[AnyContent] =>
+      withSubscribedUser(request) { status =>
+        Ok(startPage())
+      }
+  }
+
+  //TODO: how to handle the left case?
+  private def withSubscribedUser(request: RequestWithSessionDataAndRetrievedData[_])(
+    f: Either[TryingToGetIndividualsFootprint, Subscribed] => Future[Result]
```
co-category: Not sure what the left case should be?

AndyHWChung: you don't need an `Either` for `f` here in this case - you're only interested in one journey status

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/StartController.scala
AndyHWChung: would we not need to write the a `Subscribed` journey status here in session to allow for the `HomeController` to pick it up and not redirect back to start?

AndyHWChung:
LGTM

PR #67, opened 2019-10-21
Cpdp 477 iv error pages

AndyHWChung:


conf/application.conf
```diff
@@ -139,6 +144,12 @@ gg {
   origin = "cgtpd"
 }
 
+external-url {
+  gov-call-charges =  "https://www.gov.uk/call-charges"
+  additional-needs = "https://www.gov.uk/dealing-hmrc-additional-needs/overview"
+  trust-registration = "https://www.tax.service.gov.uk/trusts-forms/form/registration-of-a-trust/new"
+}
+
```
co-category: how do these urls relate to IV?

AndyHWChung: the call charges url and the additional needs url's are used for the insufficient evidence page. The trust registration url isn't used for iv - it's used to direct users to the trust registration service when they need to go and register a trust

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/IvController.scala
```diff
@@ -59,9 +75,64 @@ class IvController @Inject()(
...
+            case IvErrorStatus.FailedMatching       => Redirect(routes.IvController.getFailedMatching())
+            case IvErrorStatus.FailedIV             => Redirect(routes.IvController.getFailedIV())
+            case IvErrorStatus.InsufficientEvidence => Redirect(routes.IvController.getInsufficientEvidence())
+            case IvErrorStatus.LockedOut            => Redirect(routes.IvController.getLockedOut())
+            case IvErrorStatus.UserAborted          => Redirect(routes.IvController.getUserAborted())
+            case IvErrorStatus.Timeout              => Redirect(routes.IvController.getTimedOut())
+            case IvErrorStatus.TechnicalIssue       => Redirect(routes.IvController.getTechnicalIssue())
+            case IvErrorStatus.PreconditionFailed   => Redirect(routes.IvController.getPreconditionFailed())
+            case IvErrorStatus.Unknown(value) =>
+              logger.warn(s"Received unknown error response status from IV: $value")
```
co-category: do we not need the FailedDirectorCheck as mentioned in the design slides?

AndyHWChung: when we talked to the auth team, they said this was a status which existed but shouldn't as they never return it and are going to take it out. As the code is written it would just get swallowed up as "Unknown".

co-category: Looking at the `identity-verification-frontend` service README, we may not need to do this as this would mean verifying director info with company house. I don't think we have that as a business case

PR #79, opened 2019-11-01
cpdp-872: landing page and new sign in page with extra urls in config

nubz:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/components/text_input.scala.html
```diff
@@ -49,7 +50,7 @@
   @if(labelAsHeading) {
     <h1><label for="@{fieldId}" class="heading-xlarge">@{label}</label></h1>
   } else {
-    <label for="@{fieldId}">@{label}</label>
+    <label for="@{fieldId}" class="@{labelClass.getOrElse("")}">@{label}</label>
```
AndyHWChung: would you want `class=` a valid thing in the html if `labelClass` is not defined?

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/subscription/check_your_details.scala.html
```diff
@@ -40,8 +40,8 @@
 )
 
 @(details: SubscriptionDetails)(implicit request:Request[_], messages:Messages, appConfig: ViewConfig)
-
-@title = @{messages("subscription.title")}
+@affinity = @{details.name.fold(a => "organisation", b => "individual")}
```
AndyHWChung: this isn't broken or anything, but you could just put `_` instead of `a` and `b` if you don't care about them

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/components/text_input.scala.html
nubz: yes but scrappy - added in the default label class anyway which conveniently stops it ever being empty

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/subscription/check_your_details.scala.html
nubz: done

AndyHWChung:
LGTM

PR #82, opened 2019-11-06
cpdp-795: update subscription address now possible

AndyHWChung:


test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/address/SubscribedChangeAddressController.scala
```diff
@@ -0,0 +1,203 @@
+/*
```
co-category: File name needs to be corrected : SubscribedChangeAddressControllerSpec.scala

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/address/SubscribedChangeAddressController.scala
```diff
@@ -0,0 +1,95 @@
...
+    val updatedSubscribedDetails = journey.subscribedDetails.copy(address = address)
+
+    if(journey.subscribedDetails === updatedSubscribedDetails){
+      EitherT.pure[Future,Error](journey)
+    } else {
+      subscriptionService
+        .updateSubscribedDetails(updatedSubscribedDetails)
+        .map(_ => journey.copy(subscribedDetails = updatedSubscribedDetails))
+    }
+  }
```
co-category: what about the call to ES6? or is that separate ticket?

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/address/SubscribedChangeAddressController.scala
AndyHWChung: whoops, I@ll change that

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/address/SubscribedChangeAddressController.scala
AndyHWChung: there's a separate ticket for that

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/address/RegistrationChangeAddressController.scala
```diff
@@ -61,8 +64,10 @@ class RegistrationChangeAddressController @Inject()(
       case _                                         => Left(Redirect(controllers.routes.StartController.start()))
     }
 
-  def updateAddress(journey: RegistrationReady, address: Address): JourneyStatus =
-    journey.copy(registrationDetails = journey.registrationDetails.copy(address = address))
+  def updateAddress(journey: RegistrationReady, address: Address)(
+    implicit hc: HeaderCarrier
+  ): EitherT[Future, Error, JourneyStatus] =
+    EitherT.pure[Future, Error](journey.copy(registrationDetails = journey.registrationDetails.copy(address = address)))
```
co-category: is `pure` better here or `rightT`? I've been used `rightT`

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/address/AddressController.scala
```diff
@@ -53,7 +54,7 @@ trait AddressController[J <: JourneyStatus] {
 
   def validJourney(request: RequestWithSessionData[_]): Either[Result, (SessionData, J)]
 
-  def updateAddress(journey: J, address: Address): JourneyStatus
+  def updateAddress(journey: J, address: Address)(implicit hc: HeaderCarrier): EitherT[Future, Error, JourneyStatus]
```
co-category: should return type be:`EitherT[Future, Error, J]` rather than `EitherT[Future, Error, JourneyStatus]`

co-category: I don't think it is necessary to widen the type here

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/address/RegistrationChangeAddressController.scala
AndyHWChung: they both have the same result and it's debatable if one is better than another. `pure` let's you think of the operation in the context of a general monad whereas `rightT` is more `EitherT` specific. If you look at the implementation of `rightT` it's just an alias for `pure` - not sure why they though it was necessary to put that in

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/address/SubscribedChangeAddressController.scala
```diff
@@ -0,0 +1,95 @@
...
+      case _                                  => Left(Redirect(controllers.routes.StartController.start()))
+    }
+
+  def updateAddress(journey: Subscribed, address: Address)(
+    implicit hc: HeaderCarrier
+  ): EitherT[Future, Error, JourneyStatus] = {
+    val updatedSubscribedDetails = journey.subscribedDetails.copy(address = address)
+
+    if(journey.subscribedDetails === updatedSubscribedDetails){
+      EitherT.pure[Future,Error](journey)
```
co-category: I didn't do this for the update email - so note to myself to do that fix.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/address/AddressController.scala
AndyHWChung: I left it as `JourneyStatus` to allow for the cases where the update doesn't result in the same type of `JourneyStatus` as `J` - however, none of the implementing classes actually do that, and based on what happened with the `EmailController` having `JourneyStatus` might not work anyway. I'll change this to be `J` for now and worry about the case with a different journey type when it happens

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/address/SubscribedChangeAddressController.scala
AndyHWChung: fixed

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/address/AddressController.scala
AndyHWChung: changed

PR #83, opened 2019-11-07
CPDP-796: update contact name

co-category:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/SubscribedDetails.scala
```diff
@@ -16,9 +16,8 @@
 
 package uk.gov.hmrc.cgtpropertydisposalsfrontend.models
 
-import cats.Eq
+import cats.kernel.Eq
```
AndyHWChung: we shouldn't need to dig into the `kernel` package I don't think

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/contactname/contact_name.scala.html
```diff
@@ -31,6 +31,7 @@
 
 
 @(form: Form[ContactName],
+        amendingContactName: Boolean,
```
AndyHWChung: do we actually need this in this template? There's a similar thing for other templates because the content changes depending on whether or not you're entering it in for the first time or you're amending what you've already given us but I don' think that's the case here.

co-category: I added for the banner to indicate to the user there has been a change. Or are we using another mechanism?

AndyHWChung: the banner would appear on the homepage rather than this one - once the redirect to the homepage is done after submitting a contact name then the idea is for the banner to displayed on the homepage

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/account/manage_your_details.scala.html
```diff
@@ -71,7 +71,7 @@ <h2 class="heading-medium">
   @cyaRow(
    messages("subscription.cyaLabel.contactName"),
    Html(details.contactName.value),
-   Some(cyaChange(messages("subscription.cyaChange.contactName"), "#"))
+   Some(cyaChange(messages("subscription.cyaChange.contactName"), name.routes.SubscribedChangeContactNameController.enterContactName().url))
```
AndyHWChung: we'd only go to `SubscribedChangeContactNameController.enterContactName()` if the user registered with id. If they went through the register without id route we'd need to take them to a new controller which let them enter in an `IndividualName` - were you going to do that in a subsequent PR or something?

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/SubscribedDetails.scala
co-category: yeah, removed

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/contactname/contact_name.scala.html
co-category: fixed

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/account/manage_your_details.scala.html
co-category: :man_facepalming: missed that - will add that now

co-category: added, but see other comments

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/name/SubscribedWithoutIdChangeContactNameController.scala
```diff
@@ -0,0 +1,87 @@
...
+      EitherT.rightT[Future, Error](journey)
+    else
+      subscriptionService
+        .updateSubscribedDetails(journeyWithUpdatedName)
+        .map(_ => journey.copy(journeyWithUpdatedName))
+
+  }
+
+  override def name(journey: Subscribed): Option[IndividualName] = {
+    val Array(first, last) = journey.subscribedDetails.contactName.value.split(" ")
```
co-category: is this safe or even correct?

conf/app.routes
```diff
@@ -106,8 +106,16 @@ GET  /subscribed/amend-details/verify-email           uk.gov.hmrc.cgtpropertydis
...
-POST /subscription/amend-details/enter-contact-name      uk.gov.hmrc.cgtpropertydisposalsfrontend.controllers.name.ContactNameController.enterContactNameSubmit()
+GET  /subscription/amend-details/enter-contact-name      uk.gov.hmrc.cgtpropertydisposalsfrontend.controllers.name.SubscriptionChangeContactNameController.enterContactName()
+POST /subscription/amend-details/enter-contact-name      uk.gov.hmrc.cgtpropertydisposalsfrontend.controllers.name.SubscriptionChangeContactNameController.enterContactNameSubmit()
+
+### Change subscribed contact name
+GET  /subscribed/amend-details/enter-contact-name      uk.gov.hmrc.cgtpropertydisposalsfrontend.controllers.name.SubscribedChangeContactNameController.enterContactName()
+POST /subscribed/amend-details/enter-contact-name      uk.gov.hmrc.cgtpropertydisposalsfrontend.controllers.name.SubscribedChangeContactNameController.enterContactNameSubmit()
+
+GET  /subscribed-without-id/amend-details/enter-contact-name      uk.gov.hmrc.cgtpropertydisposalsfrontend.controllers.name.SubscribedWithoutIdChangeContactNameController.enterIndividualName()
```
co-category: is this endpoint appropriately named?

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/name/SubscribedWithoutIdChangeContactNameControllerSpec.scala
```diff
@@ -0,0 +1,104 @@
...
+      Right(IndividualName("Joe", "Smith")),
+      Email("joe.smith@gmail.com"),
+      UkAddress("21 Joe Streee", None, None, None, "BN112JJ"),
+      ContactName("Bob Smith"),
+      CgtReference("XDCGT01234568798"),
+      None,
+      false
+    )
+  )
+
```
co-category: this is a follow on comment from relying on splitting names by a single space - when I used the generator to generate SubscribedDetails it had a contact name without a space and so failed the tests...can we rely on a space always being present to split the first name and last name in the ContactName

conf/app.routes
AndyHWChung: probably not - but when we get the actual url's designed then we'll have a proper one

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/name/SubscribedWithoutIdChangeContactNameController.scala
AndyHWChung: you should be able to set this to `None` - all this is used for is to pre-populate the form with a name if one already exists which I don't think we need here

co-category: kl

AndyHWChung: but if we did want to prepopulate it we'd use the `journey.name` field instead of the `contactName` field here and look for a `Right`

co-category: replaced it with 

```scala
  override def name(journey: Subscribed): Option[IndividualName] = journey.subscribedDetails.name match {
    case Left(_)               => None
    case Right(individualName) => Some(individualName)
  }
```

AndyHWChung:
LGTM

PR #86, opened 2019-11-11
CPDP-586: update verifier details

co-category:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/SubscribedAndVerifierDetails.scala
```diff
@@ -0,0 +1,70 @@
...
+package uk.gov.hmrc.cgtpropertydisposalsfrontend.models
+
+import cats.Eq
+import play.api.libs.json.{Format, Json}
+import uk.gov.hmrc.cgtpropertydisposalsfrontend.models.EitherUtils.eitherFormat
+import uk.gov.hmrc.cgtpropertydisposalsfrontend.models.address.Address
+import uk.gov.hmrc.cgtpropertydisposalsfrontend.models.ids.CgtReference
+import uk.gov.hmrc.cgtpropertydisposalsfrontend.models.name.{ContactName, IndividualName, TrustName}
+
+final case class SubscribedAndVerifierDetails(
```
AndyHWChung: you probably could have got away with something like this:
```scala
final case class SubscriptionUpdateRequest(
  newDetails: SubscribedDetails
  previousDetails: SubscribedDetails
)
```
to avoid having all the same fields here again in this case class. We're going to have to have the old details when we come to doing the TxM audit events in the BE anyway, so we could make that work easier here.

co-category:
```scala
final case class SubscriptionUpdateRequest(
  newDetails: SubscribedDetails
  previousDetails: Option[SubscribedDetails]
)
```

The `previousDetails` would have to be optional.

AndyHWChung: would the previous details not always be at hand?

co-category: Ah, ok. I see what you mean. I was assuming we would only populate if there was an address change only. But yes if we generalise this then we could send both sets of data to the BE.

AndyHWChung: yeah. then the BE can decide whether or not to update the verifiers based on the differences it sees between the old and new details. This way, if we ever change the verifiers (which we probably will never do) only the BE will have to change and the FE can be left alone

co-category: Refactored the model + test. Need to do the the corresponding BE changes now.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/email/SubscribedChangeEmailController.scala
```diff
@@ -76,7 +76,7 @@ class SubscribedChangeEmailController @Inject()(
   ): EitherT[Future, Error, Subscribed] = {
     val journeyWithUpdatedEmail = journey.subscribedDetails.copy(emailAddress = email)
     subscriptionService
-      .updateSubscribedDetails(journeyWithUpdatedEmail)
+      .updateSubscribedDetails(SubscribedUpdateDetails(journeyWithUpdatedEmail, journey.subscribedDetails))
```
AndyHWChung: while your here it might be worth putting in the check to see if the updated details are the same as the old details if it's not too much effort

co-category: good call, I forgot about that....

co-category: fixed

AndyHWChung:
LGTM

AndyHWChung:
LGTM

PR #87, opened 2019-11-12
Cpdp 908 change notifications

AndyHWChung:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/HomeController.scala
```diff
@@ -18,40 +18,59 @@ package uk.gov.hmrc.cgtpropertydisposalsfrontend.controllers
...
     with SessionUpdates
     with Logging {
 
   def homepage(): Action[AnyContent] = authenticatedActionWithSessionData.async {
     implicit request: RequestWithSessionData[AnyContent] =>
-      withSubscribedUser(request) { subscribed =>
-        Ok(manageYourDetailsPage(subscribed.subscribedDetails))
+      withSubscribedUser(request) { (sessionData, subscribed) =>
+        val updateSessionResult =
+          sessionData.subscriptionDetailUpdated.fold[Future[Either[Error, Unit]]](
```
co-category: I'm not a `cats` expert but I believe `cats-mtl` solves these crappy type inference problems - type inference is one of the major problems with cats monad transfromers

AndyHWChung: I haven't come across the `mtl` library, but it's worth looking into if it'll make things easier. I'll have a look to see what it could give us

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/email/EmailController.scala
```diff
@@ -186,7 +188,8 @@ trait EmailController[Journey <: JourneyStatus, VerificationCompleteJourney <: J
                   _ <- EitherT[Future, Error, Unit](updateSession(sessionStore, request) { s =>
                         s.copy(
                           journeyStatus     = Some(journey),
-                          emailToBeVerified = Some(emailToBeVerified.copy(verified = true))
+                          emailToBeVerified = Some(emailToBeVerified.copy(verified = true)),
+                          subscriptionDetailUpdated = if(updateSubscriptionDetailUpdated) Some(SubscriptionDetail.Email) else None
```
co-category: `updateSubscriptionDetailUpdated` - what does this mean? it's been updated twice?

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/HomeController.scala
AndyHWChung: I'm not sure it would necessarily help in this part of the code. We're not dealing with monad transformers here - the type inference problem arises from the `fold` on the `Option`. The fold tries to derive the target type from the return type of the first `isEmpty` block. Without the type annotation it dervies `Future[Right[Nothing,Unit]]` from `Future.successful(Right(()))` and the expects that return type in the second `ifDefined` block

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/email/EmailController.scala
AndyHWChung: it was meant to mean "update the subscriptionDetailsUpated field in the session data". Wasn't quite sure what a nice name would be for that. Maybe `updateSessionSubscriptionDetailUpated`?

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/HomeController.scala
co-category: Yes, you're right in this case. I auto-jumped to cats as I was hitting that problem earlier. But this is even worse, what does that say about the Scala type system? :zipper_mouth_face:

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/email/EmailController.scala
co-category: what about `updateSubscriptionDetailChangeFlag`? bit of a mouth full but it's clear I think

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/HomeController.scala
AndyHWChung: maybe we should use our own fold on Option with better type inference :laughing:  :
```scala
  def fold[A,B](x: Option[A])(ifEmpty: => B, ifDefined: A => B): B = x.fold(ifEmpty)(ifDefined)
```

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/account/manage_your_details.scala.html
```diff
@@ -37,10 +29,11 @@
  cyaChange: uk.gov.hmrc.cgtpropertydisposalsfrontend.views.html.components.cya_change,
  addressDisplay: uk.gov.hmrc.cgtpropertydisposalsfrontend.views.html.components.address_display,
  submitButton: uk.gov.hmrc.cgtpropertydisposalsfrontend.views.html.components.submit_button,
- warningMessage: uk.gov.hmrc.cgtpropertydisposalsfrontend.views.html.components.warning
+ warningMessage: uk.gov.hmrc.cgtpropertydisposalsfrontend.views.html.components.warning,
+ alertBanner: uk.gov.hmrc.cgtpropertydisposalsfrontend.views.html.components.alert_banner
 )
 
-@(details: SubscribedDetails)(implicit request:Request[_], messages:Messages, appConfig: ViewConfig)
+@(details: SubscribedDetails, detailChanged: Option[SubscriptionDetail])(implicit request:Request[_], messages:Messages, appConfig: ViewConfig)
```
co-category: `detailChanged: Option[SubscriptionDetail])` 

should be called something like

`updatedDetails`: Option[UpdatedSubscribedDetails]`

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/email/EmailController.scala
AndyHWChung: yeah, I think that sounds better. I'll change it to that (with `change` in the past tense though I think)

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/account/manage_your_details.scala.html
AndyHWChung: I named `SubscriptionDetail` that to be as generic as possible - we could reuse for something other than indicating that a certain field has changed. It's also singluar here since you can have only changed one thing.

co-category: If you did that, then your implementation will be very confusing to read eventually. A person wouldn't know that a reference to SubscriptionDetail was to do with a change in detail or something entirely different. It's not a good approach to take.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/email/EmailController.scala
AndyHWChung: I renamed it now

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/account/manage_your_details.scala.html
co-category: Actually, the name is fine. I took another look at the defn of SubscriptionDetail

AndyHWChung: :+1:  the context should be able to tell you what it's being used for

co-category:
Have the test been run and the test coverage checked? I know Ali has issues with running both.

AndyHWChung:
yeah, I did it on my machine

PR #110, opened 2019-11-27
reduced test compilation time

AndyHWChung:


test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/StartControllerSpec.scala
```diff
@@ -1489,7 +1490,7 @@ class StartControllerSpec
                 Right(
                   Some(
                     SessionData.empty.copy(
-                      journeyStatus = Some(sample[JourneyStatus])
+                      journeyStatus = Some(sample(implicitly[ClassTag[JourneyStatus]], journeyStatusGen))
```
AndyHWChung: this is annoying but I couldn't figure out a fix. What's happening is that there's an implicit `Gen[JourneyStatus]` in scope but there are also `Gen` instances for the subtypes of `JourneyStatus` also in scope (e.g. `Gen[Subscribed]`, `Gen[SubscriptionReady]`) - the compiler is getting confused and saying that `Gen[JourneyStatus]` conflicts with those other `Gen` instances. Not sure what the best way to fix that was, but this is a workaround.

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/Generators.scala
```diff
@@ -0,0 +1,150 @@
...
+  implicit val subscribedUpdateDetailsGen: Gen[SubscribedUpdateDetails] = gen[SubscribedUpdateDetails]
+
+}
+
+trait EmailGen { this: WithGenGenerator =>
+
+  implicit val emailGen: Gen[Email] = gen[Email]
+
+  implicit val emailSourceGen: Gen[EmailSource] = gen[EmailSource]
+}
```
co-category: So now, for every new type for which we need to generate arbitrary values we will need a corresponding entry here?

co-category:
Are you able to switch from Kanji characters to something more readable? It will aid debugging alot.

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/Generators.scala
AndyHWChung: yes, the compilation time is reduced because the work that shapeless has to do to generate each `Arbitrary` instance is now once per type instead of everywhere in each file the instance was needed. It's a bit of overhead to write, but in the long run it's quicker than waiting for tests to compile

AndyHWChung:
oh yeah, I was going to look into that. I'll look at that now

co-category:
I think what kills the JSON serialisation is the presence of single/double quotes in the values generated for the fields - not sure if there is a way to exclude those characters from being used to generate values

AndyHWChung:
I've created an `Arbitrary` instance for `String` which just generates an alphachar string now which is picked up instead of the default one which generates chinese characters and quotes

PR #111, opened 2019-11-27
cpdp-1009: enforced autocomplete attribute for all user inputs

nubz:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/components/text_input.scala.html
```diff
@@ -29,7 +31,8 @@
   extraContent: Option[Html] = None,
   inputType: Option[String] = Some("text"),
   inputClass: Option[String] = None,
-  labelClass: Option[String] = None
+  labelClass: Option[String] = None,
+  autoComplete: AutoCompleteType
```
AndyHWChung: it's better to stick params without a default value before those that do. If you have this:
```scala
def f(i: Int = 1, s: String) = ???
```
you have no choice but to call `f` always supply both `i` and `s` which defeats the purpose of having the default value for `i`.

nubz: that's done

AndyHWChung:
LGTM

PR #116, opened 2019-11-28
CPDP-1002: session timeout integrated

nubz:


conf/application.conf
```diff
@@ -130,14 +130,16 @@ google-analytics {
...
-  version = "3.9.0"
+  version = "3.11.0"
   version = ${?ASSETS_FRONTEND_VERSION}
   url = "http://localhost:9032/assets/"
 }
 
 gg {
   url    = "http://localhost:9949/auth-login-stub/gg-sign-in"
   origin = "cgtpd"
+  timeout = 900
```
AndyHWChung: can we either put `-seconds` at the end of the parameter name or keep it as is but allow people to enter time units in the parameter value, e.g. `10 seconds`, `1 minute`. The second option is easier to work with - I can show you how to make that happen if you like

AndyHWChung:
LGTM

PR #124, opened 2019-12-04
cpdp-1003: set jsConfig vals individually instead of Object.assign

nubz:


nubz:
This has been tested in IE11 and IE10 and is working as expected

PR #129, opened 2019-12-06
Fix bpr audit event

co-category:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/onboarding/audit/BusinessPartnerRecordNameMatchDetails.scala
```diff
@@ -0,0 +1,35 @@
...
+
+  final case class IndividualNameWithSaUtrAuditDetails(firstName: String, lastName: String, sautr: String)
+      extends BusinessPartnerRecordNameMatchDetails
+
+  final case class TrustNameWithTrnAuditDetails(trustName: String, trn: String)
+      extends BusinessPartnerRecordNameMatchDetails
+
+  @SuppressWarnings(Array("org.wartremover.warts.PublicInference"))
+  implicit val format: OFormat[BusinessPartnerRecordNameMatchDetails] =
+    derived.oformat[BusinessPartnerRecordNameMatchDetails]
```
AndyHWChung: You may not want to used the derived format here - it will nest the JSON which we don't want. This: 
```scala
sealed trait A
case class B(i: Int) extends A
implicit val f = derived.oformat[A]
Json.toJson(B(1))
```
will result in this JSON object:
```json
{ "B" : { "i" : 1 } }
```
instead of
```json
{ "i" : 1 }
```

co-category: It's a sum type so would have to use derived format here unless there is another way?

AndyHWChung:
LGTM

PR #130, opened 2019-12-09
CPDP-1018: draft propagation of user type

co-category:
@AndyHWChung feedback on design required


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/onboarding/InsufficientConfidenceLevelController.scala
```diff
@@ -91,7 +91,10 @@ class InsufficientConfidenceLevelController @Inject()(
       InsufficientConfidenceLevelController.haveANinoForm
         .bindFromRequest()
         .fold(
-          e => BadRequest(doYouHaveANinoPage(e)),
+          e => {
+            implicit val s: Option[UserType] = request.authenticatedRequest.userType
```
AndyHWChung: to avoid having to explicitly define this implicit every time you could probably stick something like this in the controller package object:
```scala
implicit def requestToUserType(r: RequestWithSessionData[_]): Option[UserType] = 
  r.authenticatedRequest.userType
```

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/JourneyUserType.scala
```diff
@@ -0,0 +1,50 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.cgtpropertydisposalsfrontend.models
+
+import uk.gov.hmrc.cgtpropertydisposalsfrontend.models.ids.{CgtReference, GGCredId, NINO, SAUTR}
+import uk.gov.hmrc.cgtpropertydisposalsfrontend.models.onboarding.email.Email
+
+sealed trait JourneyUserType extends Product with Serializable
```
AndyHWChung: this case class is to describe the user type based on the data retrieved from auth - maybe something like RetrievedUserType would be more accurate?

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/onboarding/InsufficientConfidenceLevelController.scala
co-category: Yeah sorry, I left this hanging around - I tried three different designs one of which involved implicit conversions and this is a remnant of that

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/JourneyUserType.scala
co-category: yep, that name reflects better the semantics - I will change it.

co-category: fixed

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/onboarding/InsufficientConfidenceLevelController.scala
co-category: removed

co-category:
I've gone back to my first design choice which is to update the user-type manually in the start controller handlers. This then propagates through as expected. I have removed the session data retrieval from AuthenticatedRequest so it's back to its original impl. I think this design gives the best compromise.

co-category:
BTW, this is still in draft as the tests are broken. I'll fix those tomorrow, but in the meantime can you review this draft PR.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/onboarding/audit/BusinessPartnerRecordNameMatchDetails.scala
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.cgtpropertydisposalsfrontend.models.onboarding.audit
+
+import julienrf.json.derived
+import play.api.libs.json.OFormat
+
+sealed trait BusinessPartnerRecordNameMatchDetails extends Product with Serializable
```
AndyHWChung: isn't this already in master? This is the thing whose format instance I changed this morning

AndyHWChung:
cool, let's go with this approach - the only thing is we'll need to make sure we're definitely setting the user type when we need to in the StartController

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/onboarding/audit/BusinessPartnerRecordNameMatchDetails.scala
co-category: this branch was not forked from master - when the actual PR is created it will show all the conflicts.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/StartController.scala
```diff
@@ -69,80 +69,95 @@ class StartController @Inject()(
     with SessionUpdates
     with IvBehaviour {
 
+  @SuppressWarnings(Array("org.wartremover.warts.NonUnitStatements"))
   def start(): Action[AnyContent] = authenticatedActionWithRetrievedDataAndSessionData.async { implicit request =>
-    (
-      request.authenticatedRequest.userType,
-      request.sessionData.flatMap(_.journeyStatus)
-    ) match {
+    updateSession(sessionStore, request)(_.copy(userType = request.authenticatedRequest.userType)).flatMap {
```
AndyHWChung: do we still need this `updateSession` here if you are populating the user type in the session data in the handle methods now?

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/actions/SessionDataActionWithRetrievedData.scala
```diff
@@ -41,7 +41,9 @@ class SessionDataActionWithRetrievedData @Inject()(val sessionStore: SessionStor
   def sessionDataAction[A](
     sessionData: Option[SessionData],
     request: AuthenticatedRequestWithRetrievedData[A]
-  ): RequestWithSessionDataAndRetrievedData[A] =
-    RequestWithSessionDataAndRetrievedData(sessionData, request)
+  ): RequestWithSessionDataAndRetrievedData[A] = {
+    val data: SessionData = sessionData.getOrElse(SessionData.empty).copy(userType = request.userType)
+    RequestWithSessionDataAndRetrievedData(Some(data), request)
```
AndyHWChung: it doesn't look like `RequestWithSessionDataAndRetrievedData` needs to have `SessionData` optional in this case

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/StartController.scala
co-category: yes, for the cases where we don't go into the handlers.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/actions/SessionDataActionWithRetrievedData.scala
co-category: No it doesn't - that was your original design. Do you want to me to change it?

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/StartController.scala
AndyHWChung: the only time we don't go into a handle method is if there is session data already there - in those instances we shouldn't need to rewrite the user type

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/actions/SessionDataActionWithRetrievedData.scala
AndyHWChung: do we still need the line above it to populate the user type if the StartController is explicitly populating the user type now?

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/StartController.scala
co-category: both you comments are connected - if we don't the update in the SessionDataACtionWithRetrievedData then we need to do this update in the start controller. If we don't do the update StartController then we need to do the update in the SessionDataActionWithRetrievedData. I will take it out of the StartController.

AndyHWChung: cool - making the `SessionData` non-optional in `RequestWithSessionDataAndRetrievedData` should make some of the code later on easier as well then

AndyHWChung:
LGTM

PR #133, opened 2019-12-11
cpdp-1014: added metrics for grafana

AndyHWChung:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/metrics/Metrics.scala
```diff
@@ -0,0 +1,63 @@
...
+  val ivTechnicalIssueCounter: Counter = counter("iv.journey-status.technical-issue.count")
+
+  val ivTimeoutCounter: Counter = counter("iv.journey-status.timeout.count")
+
+  val ivUnknownErrorCounter: Counter = counter("iv.journey-status.unknown-error.count")
+  
+  val ivGetFailedJourneyStatusTimer: Timer = timer("iv.get-failed-journey-status.time")
+
+  val ivGetFailedJourneyStatusErrorCounter: Counter = counter("iv.get-failed-journey-status.errors.count")
+
```
co-category: might be worth countring how many users come through with a wrong-gg-account and also organisations which want to report a tax for a trust.

AndyHWChung: good idea, I'll add those in

co-category: also the number of users we push to register with the Trust service so we get an idea of how many times that happens

AndyHWChung: I've added in a few more metrics now for what I think it was you were describing

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/onboarding/DeterminingIfOrganisationIsTrustController.scala
```diff
@@ -88,29 +90,32 @@ class DeterminingIfOrganisationIsTrustController @Inject()(
...
-                updateSession(sessionStore, request)(
-                  _.copy(
-                    journeyStatus = Some(
-                      DeterminingIfOrganisationIsTrust(
-                        determiningIfOrganisationIsTrust.ggCredId,
-                        Some(isReportingForTrust),
-                        None
-                      )
+              BadRequest(
+                doYouWantToReportForATrustPage(formWithError, controllers.routes.StartController.weNeedMoreDetails())
```
co-category: looks like you hit return here by accident

AndyHWChung: I think that was scalafmt making the long line shorter

PR #143, opened 2020-01-02
cpdp-1068: agent enter client cgt ref page

AndyHWChung:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/StartController.scala
```diff
@@ -236,7 +241,7 @@ class StartController @Inject()(
     val newSessionData =
       request.sessionData.journeyStatus match {
         case Some(d: SubscriptionStatus.DeterminingIfOrganisationIsTrust) => d
-        case _                                                            => SubscriptionStatus.DeterminingIfOrganisationIsTrust(ggCredId, None, None)
+        case _                                                            => SubscriptionStatus.DeterminingIfOrganisationIsTrust(ggCredId, ggEmail, None, None)
       }
```
co-category: why do we need to thread in the gg email now?

AndyHWChung: this was related to the refactoring of the `StartController` I did to condense all the cases for `SubscriptionMissingData` into one case. In order to make that work I had to make sure that the gg email was always available from the session object.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/agents/AgentAccessController.scala
```diff
@@ -0,0 +1,135 @@
...
+
+  private def handleSubmittedCgtReferenceNumber(
+    cgtReference: CgtReference,
+    ggCredId: GGCredId
+  )(implicit hc: HeaderCarrier, request: RequestWithSessionData[_]): Future[Result] =
+    authorisedFunctions
+      .authorised(
+        Enrolment(CgtEnrolment.enrolmentKey)
+          .withIdentifier(CgtEnrolment.enrolmentIdentifier, cgtReference.value)
+          .withDelegatedAuthRule(CgtEnrolment.delegateAuthRule)
```
co-category: where is this rule defined? is it in the auth config?

AndyHWChung: I think it stems from the auth config here:

https://github.com/hmrc/auth/blob/19be6274975a5031e3f07005e503ec29771e6057/conf/application.conf#L260

and that references the agent-access-control service which handles agents trying to access cgt here:

https://github.com/hmrc/agent-access-control/blob/4f54211ed074cbc6fcd6fe82563cc05033b7487e/conf/app.routes#L8

PR #152, opened 2020-01-09
cpdp-1111: limit number of agent verifier match attempts

AndyHWChung:


conf/messages
```diff
@@ -449,4 +449,10 @@ agent.enter-client-country.title=Enter your client’s country
 agent.enter-client-country.extraContent=This must match the country your client used in their Capital Gains Tax on UK property account contact address.
 
 postcode.error.noMatch=Postcode cannot be found
-countryCode.error.noMatch=Country cannot be found
\ No newline at end of file
+countryCode.error.noMatch=Country cannot be found
+
+agent.too-many-attempts.title=You’ll need to wait and try again
+agent.too-many-attempts.p1=You’ve entered incorrect details for your client too many times.
+agent.too-many-attempts.p2=You’ll need to wait 2 hours and try again.
```
co-category: If this whole mechanism is designed to prevent phishing, then allowing this user to try again is bit pointless. Also when this user does exceed 10 tries, what happens then? Unless this is monitored and actioned upon, this mechanism is just giving a false sense of security IMO.

AndyHWChung:
the mechanism of the timeout is implemented with the TTL index on the mongo database which stores the unsuccessful attempts. From the controller spec point of view then the expiry is tested since the timeout will just result in nothing being returned by that store. The way to test the actual expiry is then to test that the TTL is working in `AgentVerifierRetryStore` - however because of the way TTL works in practice means that we can't really write a unit test. What seems to be happening is that a document gets some "expired" flag set on it after the configured TTL, but the record isn't actually deleted until some process cleans up expired entries. The process happens every 60 seconds by default. There are workarounds to make it happen quicker but it'll require admin access to the mongo database which may not be possible in all environments. Failing this then the best we can probably hope for is a manual test to see if it's working.

PR #155, opened 2020-01-13
CPDP-1039: Upscan Integration

co-category:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/connectors/UpscanConnector.scala
```diff
@@ -0,0 +1,141 @@
...
+          } else {
+            Right(response)
+          }
+        }
+        .recover { case e => Left(Error(e)) }
+    )
+
+  }
+
+  @SuppressWarnings(Array("org.wartremover.warts.Var", "org.wartremover.warts.Any"))
```
co-category: Wartremover is complaining about `MultipartFormData`

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/upscan/upscan.scala.html
```diff
@@ -0,0 +1,41 @@
...
+)
+
+@(
+ descriptor: UpscanDescriptor
+)(implicit request: Request[_], messages: Messages, appConfig: ViewConfig)
+
+@mainTemplate(title = "Upload file", userType = None) {
+ @form(action = routes.UpscanController.upload(), 'enctype -> "multipart/form-data"){
+ @CSRF.formField
+  <input type="hidden" name="href" value="@descriptor.uploadRequest.href" />
```
AndyHWChung: if we have the href here, someone could modify the HTML and upload something to whatever URL they put as the value here - is there a more secure way of getting the url, e.g. through mongo or something? Perhaps we include a UUID in the mongo document and submit that id instead of the url

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/UpscanController.scala
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  def errorRedirect(cgtReference: String): Action[AnyContent] = Action {
+    implicit request: MessagesRequest[AnyContent] =>
+      errorHandler.errorResult(None)
+  }
+
+  def successRedirect(cgtReference: String): Action[AnyContent] = Action {
+    implicit request: MessagesRequest[AnyContent] =>
+      Ok(upscanSuccessPage())
```
AndyHWChung: we'd never see this page rendered anywhere right since this is triggered via a callback from upscan? what responses do they expect back from the callbacks or can we reply back with anything?

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/connectors/UpscanConnector.scala
```diff
@@ -0,0 +1,141 @@
...
+  override def upload(href: String, form: MultipartFormData[Source[ByteString, _]])(
+    implicit hc: HeaderCarrier
+  ): EitherT[Future, Error, Unit] = {
+    val parts: Source[MultipartFormData.Part[Source[ByteString, _]], _] = Source.apply(form.dataParts.flatMap {
+      case (key, values) =>
+        values.map(value => MultipartFormData.DataPart(key, value): MultipartFormData.Part[Source[ByteString, _]])
+    } ++ form.files)
+    EitherT[Future, Error, Unit](
+      wsClient
```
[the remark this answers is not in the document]
co-category: `http-verbs` does have a http post method which takes a `Source[MultipartFormData.Part[Source[ByteString, _]], _]`

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/repos/upscan/UpscanStore.scala
```diff
@@ -0,0 +1,154 @@
...
+      .recover {
+        case t =>
+          logger.error(s"$message (${index.eventualName})", t)
+          false
+      }
+
+  private def ensureIndexes(implicit ec: ExecutionContext): Future[Seq[Boolean]] =
+    Future.sequence(indexes.map(ensureIndex))
+
+  ensureIndexes(scala.concurrent.ExecutionContext.Implicits.global)
```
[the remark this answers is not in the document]
co-category: not sure what you mean...index is at the collection level and this index is for the upscan collection.

What I did here was a bit hacky. I would like to have imported simple-reactivemongo but then this project would have two libraries importing reactivemongo. Mongo-caching is really horrible. I'm sure you could have achieved the same just by using simple-reactivemongo which is the recommended practice.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/upscan/upscan.scala.html
co-category: not sure what you mean here as well - I'm not sure how getting the URL out of mongo or elsewhere will make this more secure, you still have to render the form with that URL value. MITM attacks are mitigated by HTTPS/SSL.

co-category: also this network traffic is within the platform VPC

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/repos/upscan/UpscanStore.scala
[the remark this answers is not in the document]
AndyHWChung: I brought in mongo-caching to handle the TTL management for us after key-store was deprecated. I think `simple-reactive-mongo` is being brought in as a transitive dependency by mongo-caching so you might be able to use that library without having to explicitly bring it in. The `CacheMongoRepository` extends a bunch of things:

https://github.com/hmrc/mongo-caching/blob/master/src/main/scala/uk/gov/hmrc/cache/repository/CacheRepository.scala#L46

One of the things it extends is something called `TTLIndexing`:

https://github.com/hmrc/mongo-caching/blob/master/src/main/scala/uk/gov/hmrc/cache/repository/TTLIndexing.scala#L28

That also does things with the indices and I just wanted to double check that that behaviour wasn't being overwritten by what you were doing here or anything like that.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/connectors/UpscanConnector.scala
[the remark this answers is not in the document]
AndyHWChung: this is an annoying limitation of `http-verbs` - what it means that is we lose all the stuff http verbs does whenever we do a http request like this:
 
https://github.com/hmrc/http-verbs/blob/master/src/main/scala/uk/gov/hmrc/http/HttpPost.scala#L42

that does some automatic logging for us when we do the request. Ideally they'd have an extra `POST` method which implicitly requires the body to be have a `BodyWritable` instance. Failing that we would have to create a `Writes` instance for one of `Source[MultipartFormData.Part[Source[ByteString, _]], _]`, `BodyWritable` or `WsBody` which might not be straightforward. 

c.f. https://github.com/playframework/playframework/blob/ddf3a7ee4285212ec665826ec268ef32b5a76000/transport/client/play-ws/src/main/scala/play/api/libs/ws/WSBodyWritables.scala#L16

The post method you're using on wsClient is also deprecated

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/upscan/upscan.scala.html
AndyHWChung: is this html page ever displayed to the user on the browser?

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/connectors/UpscanConnector.scala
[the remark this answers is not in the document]
co-category: I don't see any deprecation annotation

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/upscan/upscan.scala.html
co-category: ofc - it's the page which the user sees to choose the file and upload it

co-category: it needs UX'ing by the UX team and then styled by Ali

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/UpscanController.scala
co-category: it depends - the UX team need to design this user flow.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/connectors/UpscanConnector.scala
```diff
@@ -0,0 +1,141 @@
...
+import play.mvc.Http.Status
+import uk.gov.hmrc.cgtpropertydisposalsfrontend.controllers.routes
+import uk.gov.hmrc.cgtpropertydisposalsfrontend.http.HttpClient._
+import uk.gov.hmrc.cgtpropertydisposalsfrontend.models.Error
+import uk.gov.hmrc.cgtpropertydisposalsfrontend.models.ids.CgtReference
+import uk.gov.hmrc.cgtpropertydisposalsfrontend.util.Logging
+import uk.gov.hmrc.http.{HeaderCarrier, HttpResponse}
+import uk.gov.hmrc.play.bootstrap.http.HttpClient
+
+import scala.concurrent.ExecutionContext.Implicits.global
```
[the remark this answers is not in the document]
co-category: fixed

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/UpscanController.scala
```diff
@@ -0,0 +1,148 @@
...
+import uk.gov.hmrc.cgtpropertydisposalsfrontend.repos.SessionStore
+import uk.gov.hmrc.cgtpropertydisposalsfrontend.services.UpscanService
+import uk.gov.hmrc.cgtpropertydisposalsfrontend.services.UpscanService.UpscanNotifyResponse
+import uk.gov.hmrc.cgtpropertydisposalsfrontend.services.UpscanService.UpscanServiceResponse._
+import uk.gov.hmrc.cgtpropertydisposalsfrontend.services.onboarding.SubscriptionService
+import uk.gov.hmrc.cgtpropertydisposalsfrontend.util.Logging
+import uk.gov.hmrc.cgtpropertydisposalsfrontend.views._
+import uk.gov.hmrc.play.bootstrap.controller.FrontendController
+
+import scala.concurrent.ExecutionContext.Implicits.global
```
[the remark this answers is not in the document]
co-category: fixed

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/UpscanController.scala
```diff
@@ -0,0 +1,148 @@
...
+        href =>
+          val data: Map[String, Seq[String]] = multipart.dataParts.map {
+            case (k, v) => (k.stripPrefix("upscan."), v)
+          }
+          val withSource =
+            multipart.files.map(file => file.copy(ref = FileIO.fromPath(file.ref.path): Source[ByteString, Any]))
+          val prepared = multipart.copy(files = withSource, dataParts = data)
+          upscanConnector.upload(href, prepared).value.map {
+            case Left(error) =>
+              logger.warn(s"Could not upload file to S3 due to $error")
```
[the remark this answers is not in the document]
co-category: fixed

conf/app.routes
```diff
@@ -12,4 +11,11 @@ GET /sign-out-and-log-in                                                    uk.g
 GET /exit/we-signed-you-out                                                      uk.gov.hmrc.cgtpropertydisposalsfrontend.controllers.StartController.timedOut()
 GET /keep-alive                                                             uk.gov.hmrc.cgtpropertydisposalsfrontend.controllers.StartController.keepAlive()
 
+# UPSCAN
```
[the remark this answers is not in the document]
co-category: created a new upscan package and put the controller inside there

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/UpscanController.scala
```diff
@@ -0,0 +1,148 @@
...
+  def upload(): Action[MultipartFormData[Files.TemporaryFile]] = Action(parse.multipartFormData(maxFileSize)).async {
+    implicit request: MessagesRequest[MultipartFormData[Files.TemporaryFile]] =>
+      val multipart = request.body
+      multipart.dataParts.get("href").flatMap(_.headOption).fold(Future.successful(BadRequest("missing href"))) {
+        href =>
+          val data: Map[String, Seq[String]] = multipart.dataParts.map {
+            case (k, v) => (k.stripPrefix("upscan."), v)
+          }
+          val withSource =
+            multipart.files.map(file => file.copy(ref = FileIO.fromPath(file.ref.path): Source[ByteString, Any]))
```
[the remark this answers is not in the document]
co-category: that's the API contract - akka-http represents a file uploaded using multipart/form-data encoding as Source[ByteString, Any] - not everything in world is well typed :stuck_out_tongue:

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/repos/upscan/UpscanStore.scala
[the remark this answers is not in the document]
co-category: Given what you said earlier today about a draft-id which will exist in the future, then that could be used to represent the id from the uploaded files which means that mongo-caching can be used as is and this indexing code can be removed. When the development reaches that point, I suggest we revisit this (as well storing the S3 file descriptor info) at that point

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/UpscanController.scala
AndyHWChung: the callback is asynchronous so these pages would never render - I imagine when we come to do the file upload in our journey we'd make the callback update some flag in mongo somewhere which will indicate that the file upload was successful. I'm imagining we'd have something like:
```scala
case class AttachedFile(id: ID, upscanSuccess: Option[Boolean])
case class DraftReturn(..., attachedFiles: List[AttachedFile])
```
with draft returns stored somewhere.

AndyHWChung: what you have now isn't broken and we can't really do anything properly with the callbacks atm since we don't have a draft return store or anything. For now though it'd be useful to see a log message if it was a success or failure with the body of the request body logged just so we can see what's happening when we test it.

AndyHWChung: oh wait a sec, this isn't the callback. Ignore me :D

co-category: :smiley: - the upscan service is badly designed tho - browser redirects :man_facepalming:

co-category: I have reverted back to v1 initiate. This redirect feature of upscan is no good and we don't really need it. V1 allows us to handle success/error statuses from within the CGT service. Also with v2, there would be no way to block users from uploading files greater than 5 MB unless client side JS is added. For users who don't have JS enabled in their browser this could potentially lead to DMS failures.

AndyHWChung:
LGTM

PR #157, opened 2020-01-14
cpdp-1112: invidual user type triage question

AndyHWChung:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/AddressController.scala
```diff
@@ -86,7 +86,7 @@ trait AddressController[J <: JourneyStatus] {
             updateSession(sessionStore, request)(_.copy(addressLookupResult = None)).map {
               case Left(e) =>
                 logger.warn(s"Could not clear addressLookupResult", e)
-                errorHandler.errorResult(request.userType)
```
co-category: Why has this been removed? If I remember what Ali said, this is required.

co-category: Why has this been removed? If it is not required to report the userType in an error scenario then we should change the function definition of errorResult so that it does not have a userType parameter.

https://github.com/hmrc/cgt-property-disposals-frontend/blob/71b83d275d7b7de5d6a3863ae8b0de2c9f5541a2/app/uk/gov/hmrc/cgtpropertydisposalsfrontend/config/ErrorHandler.scala#L38

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/accounts/homepage/HomePageController.scala
```diff
@@ -0,0 +1,91 @@
...
+
+  private def withSubscribedUser(request: RequestWithSessionData[_])(
+    f: (SessionData, Subscribed) => Future[Result]
+  ): Future[Result] =
+    request.sessionData.flatMap(s => s.journeyStatus.map(s -> _)) match {
+      case Some((s: SessionData, r: Subscribed)) =>
+        f(s, r)
+      case _ =>
+        Future.successful(SeeOther(uk.gov.hmrc.cgtpropertydisposalsfrontend.controllers.routes.StartController.start().url))
+    }
```
co-category: import the controller package to make the line shorter - same elsewhere in this file

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/AddressController.scala
AndyHWChung: I created an overloaded method which didn't take an `Option[UserType]` and instead took an implicit `RequestWithSessionData[_]` so that we could skip writing `errorHandler.errorResult(request.userType)` everywhere since the user type can be dervied easily from a `RequestWithSessionData[_]`

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/accounts/homepage/HomePageController.scala
```diff
@@ -0,0 +1,91 @@
...
+                                val sessionDataAction: SessionDataAction,
+                                errorHandler: ErrorHandler,
+                                sessionStore: SessionStore,
+                                cc: MessagesControllerComponents,
+                                manageYourDetailsPage: views.html.account.manage_your_details,
+                                homePage: views.html.account.home,
+                                privateBetaHomePage: views.html.account.home_private_beta,
+                                detailUpdatedPage: views.html.account.details_updated,
+                                signedOutPage: views.html.account.signed_out
+                              )(implicit viewConfig: ViewConfig, ec: ExecutionContext)
```
co-category: apply the formatting - this block of parameters should be two spaces from the left I think

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/Generators.scala
```diff
@@ -43,8 +44,9 @@ object Generators
     with NameMatchGen
     with OnboardingDetailsGen
     with EmailGen
+    with VerifierMatchGen
     with UserTypeGen
-    with VerifierMatchGen {
+    with TriageQuestionsGen {
```
co-category: By Triage, do we mean ascertaining the eligibility to use the service?

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/accounts/homepage/HomePageController.scala
AndyHWChung: done

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/accounts/homepage/HomePageController.scala
AndyHWChung: and done - I never figured out why the formatter wasn't doing anything on compile

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/Generators.scala
AndyHWChung: basically yeah - there'll be a bunch of exit points in the triage section where the user will have to do something else based on their situation

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/AddressController.scala
AndyHWChung: see this: https://github.com/hmrc/cgt-property-disposals-frontend/pull/157/files#diff-8e4beadcc80d3be708ac788c111a3ed4R47

AndyHWChung: don't know if that's going to take you to the right place actually - but it's supposed to take you to the diff for `ErrorHandler.scala`

PR #158, opened 2020-01-14
cpdp-1152: unsuccssful verifier match count is not increased when same value is submitted

AndyHWChung:
…e value is submitted


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/agents/AgentAccessController.scala
```diff
@@ -193,25 +194,28 @@ class AgentAccessController @Inject()(
...
+              journeyStatus = Some(
+                Subscribed(verifierMatchingDetails.clientDetails, agentSupplyingClientDetails.agentGGCredId)
+              )
+            )
+          ).map {
             case Left(e) =>
-            logger.warn("Could not update session", e)
-            errorHandler.errorResult(request.userType)
+              logger.warn("Could not update session", e)
+              errorHandler.errorResult(request.userType)
```
co-category: `errorHandler.errorResult(request.userType)` -> `errorHandler.errorResult()` as per #157

AndyHWChung: I've changed that now

PR #163, opened 2020-01-20
cpdp-1106: triage question disposal date

AndyHWChung:
fixed user type population for subscribed users


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/triage/CanTheyUseOurServiceController.scala
```diff
@@ -120,36 +130,91 @@ class CanTheyUseOurServiceController @Inject()(
...
+  def whenWasDisposalDate(): Action[AnyContent] = authenticatedActionWithSessionData.async { implicit request =>
+    withIndividualTriageAnswers(request) {
+      case (_, subscribed, individualTriageAnswers) =>
+        if (individualTriageAnswers.numberOfProperties.isEmpty) {
+          Redirect(routes.CanTheyUseOurServiceController.howManyProperties())
+        } else {
+          val today = LocalDate.now(clock)
+          val form = subscribed.individualTriageAnswers
+            .flatMap(_.disposalDate)
+            .fold(disposalDateForm(today))(disposalDateForm(today).fill)
```
co-category: `today` is UTC
How do you know what timezone the user is in? Say they are overseas residents.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/triage/CanTheyUseOurServiceController.scala
```diff
@@ -120,36 +130,91 @@ class CanTheyUseOurServiceController @Inject()(
...
+    }
+  }
+
+  def whenWasDisposalDate(): Action[AnyContent] = authenticatedActionWithSessionData.async { implicit request =>
+    withIndividualTriageAnswers(request) {
+      case (_, subscribed, individualTriageAnswers) =>
+        if (individualTriageAnswers.numberOfProperties.isEmpty) {
+          Redirect(routes.CanTheyUseOurServiceController.howManyProperties())
+        } else {
+          val today = LocalDate.now(clock)
```
co-category: Better to use `Instant` to represent a moment in time on the UTC timeline.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/triage/CanTheyUseOurServiceController.scala
AndyHWChung: I had a chat with Jo and Ash about this from a BA and UR angle. The conclusion was came to was that it's probably best to measure today against the "Europe/London" timezone. Putting it to the JVM's timezone wouldn't work because it would likely set it to the timezone the data centre hosting our service is in.

To use our service you need to have completed the disposal - so the completion date has to be "today" or before. The disposal date must be before the completion date so that must also be today or before. We went through this scenario:
 - completion date is set for 2nd Jan
 - user is in Australia somewhere where the time difference is 12 hours. It is 11pm 1st Jan in the UK and it is 11am 2nd Jan for the user in Australia
- user tries to use our service to start a return
- user enters in their "today" for the completion date (2nd Jan)
- we see that the date is before our today (1st Jan) and display a message which currently says something like "completion date must be today or before"

This would probably cause confusion for the user, so we're going to double check the error message here. But in this scenario, the completion could not have occurred at the time the user was doing it since the disposal must be done in the UK, so no-one would have actually done any of the paperwork necessary for the completion to take place in the scenario above. The above scenario is very unlikely to happen unless someone is super eager to do their return.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/triage/CanTheyUseOurServiceController.scala
AndyHWChung: we don't need/have any time information so that part of `Instant` is wasted on us a bit. We only ever use the date as a day, month, year combo so I'm not sure what more `Instant`  would give us

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/triage/CanTheyUseOurServiceController.scala
AndyHWChung: I changed the clock to be `Clock.system(ZoneId.of("Europe/London"))` now

co-category: No, the JVM TZ will be set according to the configuration of the VM the JVM lives in. I checked with the infrastructure team and they set it to UTC. This will only ever come an issue if you start doing date calculations inside the CGT service which I don't see anywhere at the moment. But the correct thing to do for now is to be consistent in the timezones. 

Given that the JVM TZ is UTC, all dates entered by the user will be parsed as UTC dates as you are using LocalDate, even if the user may consider the date he is entering to be of some other TZ. So the the `today` value should also be

```scala
val today = LocalDate.now(ZoneOffset.UTC)
```

or even just

```scala
val today = LocalDate.now
```
if you trust the platform.

AndyHWChung: I wouldn't trust the platform - I'd still stick with explicitly stating UTC in our code to make sure. I'll revert to what I had when I created this PR then and use `Clock.systemUTC()` to make sure the LocalDate is always relative to UTC

AndyHWChung: I've changed it back to UTC now

AndyHWChung:
I've justed removed `returns.routes` from `private_beta.routes` since we'll never need the returns stuff in private beta

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/triage/CanTheyUseOurServiceControllerSpec.scala
```diff
@@ -431,36 +463,387 @@ class CanTheyUseOurServiceControllerSpec
...
+            // month invalid and takes precedence over year
+            (Some("1"), Some("13"), Some("2020"), "disposalDate-month.error.invalid"),
+            (Some("1"), Some("0"), Some("0"), "disposalDate-month.error.invalid"),
+            (Some("1"), Some("-1"), Some("-6"), "disposalDate-month.error.invalid"),
+            (Some("1"), Some("1.2"), Some("3.4"), "disposalDate-month.error.invalid"),
+            // year invalid
+            (Some("1"), Some("12"), Some("0"), "disposalDate-year.error.invalid"),
+            (Some("1"), Some("12"), Some("-1"), "disposalDate-year.error.invalid"),
+            (Some("1"), Some("12"), Some("1.2"), "disposalDate-year.error.invalid"),
+            // date in the future
```
co-category: I've not explored it but can generators be used here to generate all the permutations?

AndyHWChung: getting the permutations of the input values shouldn't be too difficult but it gets difficult when you try to pairing each permutation with a specific error message. I expect that would take some complex logic to write which will end up looking quite similar to the actual production code which decides the error message

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/triage/CanTheyUseOurServiceController.scala
co-category: I'd prefer something like this as it makes it clear what's going on and dare I say, more idiomatic in terms of the usage os the Java time API:
This is what I meant but leave it as is. This encoding, for me, makes it clear what the intent is but YMMV.

```scala
Instant instantOfNow = Instant.now();
LocalDate localDate = LocalDateTime.ofInstant(instantOfNow, ZoneOffset.UTC).toLocalDate();
```

PR #165, opened 2020-01-22
cpdp-871: handle case where bpr call indicates user is already subscribed

AndyHWChung:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/actions/AuthenticatedActionWithRetrievedData.scala
```diff
@@ -175,7 +175,7 @@ class AuthenticatedActionWithRetrievedData @Inject()(
         }
       case None =>
         subscriptionService
-          .hasSubscription()
+          .hasFailedCgtEnrolment()
```
AndyHWChung: I thought this was a more accurate name

PR #166, opened 2020-01-23
CPDP-1040: integrate PDF generator service

co-category:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/services/PdfGeneratorService.scala
```diff
@@ -0,0 +1,43 @@
...
+trait PdfGeneratorService {
+  def generatePdfBytes(html: String)(implicit hc: HeaderCarrier): EitherT[Future, Error, Stream[Char]]
+}
+
+@Singleton
+class PdfGeneratorServiceImpl @Inject() (pdfGeneratorConnector: PdfGeneratorConnector)
+    extends PdfGeneratorService
+    with Logging {
+
+  override def generatePdfBytes(html: String)(implicit hc: HeaderCarrier): EitherT[Future, Error, Stream[Char]] = {
```
co-category: will be called from within the module to do the DMS submission

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/connectors/pdf/PdfGeneratorConnector.scala
```diff
@@ -0,0 +1,64 @@
...
+@Singleton
+class PdfGeneratorConnectorImpl @Inject() (config: ServicesConfig, http: HttpClient)(implicit ec: ExecutionContext)
+    extends PdfGeneratorConnector
+    with Logging {
+
+  override def generatePDF(payload: Map[String, Seq[String]], headers: Seq[(String, String)])(
+    implicit hc: HeaderCarrier
+  ): EitherT[Future, Error, Stream[Char]] = {
+
+    val serviceName  = "pdf-generator"
```
AndyHWChung: these `vals` can sit outside of the `def` - saves allocation time everytime we call this function

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/connectors/pdf/PdfGeneratorConnector.scala
```diff
@@ -0,0 +1,64 @@
...
+    lazy val baseURL = config.baseUrl(serviceName) + config.getConfString(s"$serviceName.base-path", "")
+
+    val url = s"$baseURL/pdf-generator-service/generate"
+
+    EitherT[Future, Error, Stream[Char]](
+      http
+        .POSTForm(url, payload, headers)
+        .map[Either[Error, Stream[Char]]] { response =>
+          if (response.status != Status.OK) {
+            Left(Error("Could not generate PDF"))
```
AndyHWChung: having the body of the error response and error status code would be useful info in the error message

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/services/PdfGeneratorService.scala
AndyHWChung: I imagine this would be in the BE service just after we've got a successful response from submitting a return - it might make more sense to stick this pdf service integration in the BE rather than the FE.The generation and the sending to DMS could be part of the submit return endpoint the BE exposes.

AndyHWChung:
is there any way to test this? Could we for example have a dummy page in the `testOnly` package or something which will trigger the pdf generation and somehow let us download what has been generated to inspect it?

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/services/PdfGeneratorService.scala
co-category: Yes, agreed.

co-category:
moved to BE service

PR #167, opened 2020-01-23
cpdp-840-728-1147: implemented further triage pages

AndyHWChung:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/returns/IndividualTriageAnswers.scala
```diff
@@ -21,12 +21,15 @@ import play.api.libs.json.{Json, OFormat}
...
+  disposalMethod: Option[DisposalMethod],
+  wasAUKResident: Option[Boolean],
+  wasResidentialProperty: Option[Boolean],
   disposalDate: Option[DisposalDate]
 )
 
 object IndividualTriageAnswers {
 
-  val empty: IndividualTriageAnswers = IndividualTriageAnswers(None, None, None)
+  val empty: IndividualTriageAnswers = IndividualTriageAnswers(None, None, None, None, None, None)
```
co-category: So this will grow linearly with the number of triage questions? Isn't there 80+ questions which will break the case class limit for number of fields?

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/testonly/JourneyStatusController.scala
```diff
@@ -137,14 +165,41 @@ object JourneyStatusController {
...
-        IndividualTriageAnswers(individualUserType, numberOfProperties, disposalDate.map(DisposalDate(_)))
+      case (
+          individualUserType,
+          numberOfProperties,
+          disposalMethod,
+          wasAUKResident,
+          disposedOfResidentialProperty,
+          disposalDate
+          ) =>
+        // TODO: fix
```
co-category: remove if not required anymore

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/returns/IndividualTriageAnswers.scala
AndyHWChung: For the triage questions there's going to be a limited number of questions so this shouldn't get massively big. Also, different states within the triage section may get their own case class depending on how it pans out. The whole return itself will have loads of questions in it, but it should be described using more than one case class

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/testonly/JourneyStatusController.scala
AndyHWChung: whoops, removed

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/returns/IndividualTriageAnswers.scala
co-category: OK, let's see how to unfolds...

PR #170, opened 2020-01-23
cpdp-933: implemented completion date triage page

AndyHWChung:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/triage/CanTheyUseOurServiceController.scala
```diff
@@ -378,9 +399,10 @@ object CanTheyUseOurServiceController {
 
   def disposalDateForm(maximumDateInclusive: LocalDate): Form[DisposalDate] = Form(
     mapping(
-      "date" -> of(
+      "" -> of(
```
co-category: how can this be an empty string?

AndyHWChung: the actual mapping is inside the formatter, so it doesn't matter what this particular string is

PR #172, opened 2020-01-24
CPDP-1156: upgrade to sbt 1.3.7 & scala 2.12

co-category:


test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/actions/AuthenticatedActionWithRetrievedDataSpec.scala
```diff
@@ -115,7 +115,7 @@ class AuthenticatedActionWithRetrievedDataSpec
         val result = performAction(FakeRequest())
 
         status(result)        shouldBe OK
-        contentAsJson(result) shouldBe Json.toJson(RetrievedUserType.NonGovernmentGatewayRetrievedUser(providerType))
+        contentAsJson(result) shouldBe Json.parse("""{"NonGovernmentGatewayRetrievedUser":{"authProvider":"other provider"}}""")
```
co-category: can you review this changes to the tests in this file

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/testonly/JourneyStatusController.scala
```diff
@@ -202,7 +202,7 @@ object JourneyStatusController {
     }
   )
 
-  implicit class MapOps[A, B](val m: Map[A, B]) extends AnyVal {
+  implicit class MapOps[A, B](private val m: Map[A, B]) extends AnyVal {
```
co-category: Scalafix injected this.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/RetrievedUserType.scala
```diff
@@ -29,27 +32,55 @@ object RetrievedUserType {
...
+    implicit val format: OFormat[IndividualWithInsufficientConfidenceLevel] = derived.oformat()
+  }
+
   final case class Subscribed(
     cgtReference: CgtReference,
     ggCredId: GGCredId
   ) extends RetrievedUserType
 
+  object Subscribed{
+    implicit val format: OFormat[Subscribed] = derived.oformat()
```
AndyHWChung: you don't need to use the derived instance here: you can just use the straightforward `Json.format` - the dervied instances are useful when you want to get a format instance for a whole sealed trait and you can't be bothered to write a big pattern match. 
This is why the `AuthenticatedActionWithRetrievedDataSpec` was failing for you. Do you even need the format instances? These are never serialised except for in that spec.

AndyHWChung:
I get a bunch of compile time warnings when I run `test:compile` - some of those look like genuine things that can be fixed, e.g some unused imports

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/ContactNameController.scala
```diff
@@ -22,7 +22,7 @@ import play.api.mvc._
 import uk.gov.hmrc.cgtpropertydisposalsfrontend.config.{ErrorHandler, ViewConfig}
 import uk.gov.hmrc.cgtpropertydisposalsfrontend.controllers.actions.{RequestWithSessionData, WithAuthAndSessionDataAction}
 import uk.gov.hmrc.cgtpropertydisposalsfrontend.models.name.ContactName
-import uk.gov.hmrc.cgtpropertydisposalsfrontend.models.onboarding.SubscriptionDetail
+//import uk.gov.hmrc.cgtpropertydisposalsfrontend.models.onboarding.SubscriptionDetail
```
AndyHWChung: you can probably delete this right?

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/RetrievedUserType.scala
co-category: No it even failed when using the Json.format. I switched to derived instances as I thought it would be clever enough to generate the correct JSON but it didn't. I can take these out now as I changed the tests to manually construct the JSON

AndyHWChung: the derived format will include the case class/object name in the JSON which isn't normally there when you use `Json.format`

co-category: The compiler complained that it could not find codecs for the these types in the test. That's why I added them. The tests seemed broken before the upgrade but it wasn't getting picked up.

co-category:
> I get a bunch of compile time warnings when I run `test:compile` - some of those look like genuine things that can be fixed, e.g some unused imports

I don't get any. Tests compile and run fine.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/ContactNameController.scala
co-category: Yes, I was testing the scalafix feature.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/RetrievedUserType.scala
co-category: > the derived format will include the case class/object name in the JSON which isn't normally there when you use `Json.format`

it didn't do that when I tried  - you can try and change the code back and see if it works for you

AndyHWChung:
LGTM

PR #175, opened 2020-01-27
CPDP-1156: workaround for scalafmt bug and format all sources

mikail-khan:
@AndyHWChung This will fix the formatting issue on compile time.

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -47,5 +47,5 @@ spaces {
 }
 
 project.excludeFilters = [
-  ".*\\.sc"
+  ".*\\.sc$"
 ]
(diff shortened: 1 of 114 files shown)


.scalafmt.conf
```diff
@@ -47,5 +47,5 @@ spaces {
 }
 
 project.excludeFilters = [
-  ".*\\.sc"
+  ".*\\.sc$"
```
mikail-khan: This was preventing the sources from being formatted on compile time. There is a bug in Scalafmt which read that reg exp incorrectly. Until they fix the bug, this is the workaround. We should also add, css, js, and html extensions here but I will do that in another PR.

AndyHWChung: Great, finality it will work!

PR #217, opened 2020-02-27
CPDP-1169 Extend test coverage.

rszmacinski:


test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/CheckAllAnswersAndSubmitControllerSpec.scala
```diff
@@ -66,6 +67,7 @@ class CheckAllAnswersAndSubmitControllerSpec
     )
 
   lazy val controller = instanceOf[CheckAllAnswersAndSubmitController]
+  lazy val conf = instanceOf[ViewConfig]
```
AndyHWChung: you don't need this `val` - you should just be able to use `viewConfig` where you need it

PR #222, opened 2020-03-02
CPDP-1114: Change to numberOfProperties

rambabu-posa:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/triage/MultipleDisposalsTriageController.scala
```diff
@@ -62,25 +67,55 @@ class MultipleDisposalsTriageController @Inject() (
...
+    withMultipleDisposalTriageAnswers(request) {
+      case (_, _, _) =>
+        numberOfPropertiesForm
+          .bindFromRequest()
+          .fold(
+            formWithErrors =>
+              BadRequest(
+                howManyProperties(formWithErrors)
+              ), { numberOfProperties =>
+              if (numberOfProperties > 1)
```
AndyHWChung: here you need to update the session data with the new answer - if the journey is a `FillingOutReturn` you need to update the draft return as well. Otherwise, if it's a `StartingNewDraftReturn` you need to make the new draft return from multile to single if the answer is one.

conf/messages
```diff
@@ -969,4 +956,12 @@ returns.list.noTaxDue=You did not owe any tax for this return
 returns.list.leftToPay=Total left to pay
 returns.list.taxOwed=Tax owed for return
 returns.list.sentDate=Sent to HMRC on {0}
-returns.paymentTakes3Days=Payments may take up to 3 working days to be shown on this page
\ No newline at end of file
+returns.paymentTakes3Days=Payments may take up to 3 working days to be shown on this page
+
+multiple-disposals.howManyProperties.title=How many disposals are you including in this return?
+multiple-disposals.howManyProperties.p1=The property disposals must all have the same 'completion' date. The contract 'exchange' dates must all be in the same tax year.
```
AndyHWChung: we use curly quotes like the ones here https://smartquotesforsmartpeople.com/ instead of straight quotes. You can copy and paste the right and left quotes from that page

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/triage/MultipleDisposalsTriageController.scala
```diff
@@ -62,37 +69,104 @@ class MultipleDisposalsTriageController @Inject() (
...
       case _ =>
         Redirect(uk.gov.hmrc.cgtpropertydisposalsfrontend.controllers.routes.StartController.start())
     }
 
 }
+object MultipleDisposalsTriageController {
+
+  val numberOfPropertiesForm: Form[Int] = Form(
+    mapping(
+      "numberOfProperties" -> of[Int]
```
AndyHWChung: this should be `text` instead of `of[Int]` and then checking that
- the string is not empty
- the string can be converted to an Int 
- the Int is strictly greater than zero and strictly less than 1000
The error messages for those can be found in the ticket

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/testonly/JourneyStatusController.scala
```diff
@@ -119,7 +119,7 @@ class JourneyStatusController @Inject() (
   ): IncompleteSingleDisposalTriageAnswers =
     IncompleteSingleDisposalTriageAnswers(
       answers.individualUserType,
-      answers.numberOfProperties,
+      answers.numberOfProperties.isDefined,
```
AndyHWChung: it more accurate to have `answers.numberOfProperties.contains(NumberOfProperties.One)` here

AndyHWChung:
LGTM

PR #248, opened 2020-03-09
CPDP-1114: cpdp-1114-multi-disposals-how-many-props

rambabu-posa:


conf/messages
```diff
@@ -1009,6 +1009,8 @@ viewReturn.summaryHeading=Your return
 
 multiple-disposals.triage.caption=Reporting multiple disposals
 
+multiple-disposals.triage.caption=Reporting multiple disposals
```
AndyHWChung: this looks like a duplicate of the message above

PR #264, opened 2020-03-12
Cpdp 1323 capacitor personal rep page

AndyHWChung:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/triage/capcitors_personal_representatives_not_handled.scala.html
```diff
@@ -0,0 +1,107 @@
...
+  @messages(s"$key.openingTimes.sunBankHols")
+ </p>
+
+  <p>
+   @Html(messages(s"$key.getHelp", appConfig.additionalNeedsUrl))
+  </p>
+
+  <p>
+   <a href="@{appConfig.callChargesUrl}">
+    @messages(s"$key.callChargesLink")
```
modivilioglu: could we not do this like above ?
@Html(messages(s"$key.callChargesLink", appConfig. callChargesUrl))

AndyHWChung: done - I think the reason the two was different is because the first link is a link mid-sentence but the second link was not

PR #280, opened 2020-03-16
[CPDP-948] - Fixed cosmetic bug and updated the tests

modivilioglu:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/initialgainorloss/initial_gain_or_loss.scala.html
```diff
@@ -45,7 +45,7 @@
         content = Some(unitInput(
             fieldId = gainKey,
             value = form.data.getOrElse(gainKey, ""),
-            label = Html(messages(s"$gainKey.label")),
+            label = Html(""),
```
AndyHWChung: do we really not need a label anymore? Doesn't the screenshot of it have one?

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/initialgainorloss/initial_gain_or_loss.scala.html
```diff
@@ -61,7 +61,7 @@
         content = Some(unitInput(
             fieldId = lossKey,
             value = form.data.getOrElse(lossKey, ""),
-            label = Html(messages(s"$lossKey.label")),
+            label = Html(""),
```
rszmacinski: I'm not sure about this, should it be empty string ?

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/initialgainorloss/initial_gain_or_loss.scala.html
modivilioglu: you are right. my bad :/ I've also added test cases to it.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/initialgainorloss/initial_gain_or_loss.scala.html
modivilioglu: you are right. my bad :/ I've also added test cases to it.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/initialgainorloss/initial_gain_or_loss.scala.html
modivilioglu: Done.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/initialgainorloss/initial_gain_or_loss.scala.html
modivilioglu: Done.

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/initialgainorloss/InitialGainOrLossControllerSpec.scala
```diff
@@ -139,6 +139,8 @@ class InitialGainOrLossControllerSpec
             ), { doc =>
               doc.select(".govuk-caption-xl").html()  should include("Initial gain or loss")
               doc.select("#initialGainOrLoss").html() should include("Did you make an initial gain or loss?")
+              doc.select("#initialGainOrLoss").html() should include("Initial gain amount")
+              doc.select("#initialGainOrLoss").html() should include("Initial loss amount")
```
AndyHWChung: this is the same thing twice. Also, is it possible to have a stronger test which has `shouldBe` instead of `should include`? if you do `doc.select("...").text()` would that work?

modivilioglu: Ok, one was for loss and the other one was for gain. but sure.

AndyHWChung: oh ok, I see that now. What would make it easier to test then is using a more specific selector or perhaps giving the different elements different id's to be able to assert that a precise element has a particular text

modivilioglu: Yes sure, I'll do that.

PR #284, opened 2020-03-18
added extra check for agent on reporting for page

ugh92:


test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/triage/CommonTriageQuestionsControllerSpec.scala
```diff
@@ -252,6 +252,7 @@ class CommonTriageQuestionsControllerSpec
                     doc.select("#individualUserType > div:nth-child(3) > label").text() shouldBe messageFromMessageKey(
                       s"individualUserType.agent.${IndividualUserType.PersonalRepresentative}"
                     )
+                    doc.body().text() shouldNot include(s"individualUserType.agent.${IndividualUserType.Capacitor}")
```
AndyHWChung: shouldn't you wrap the key in `messageFromMessageKey` to make sure that your'e escaping the key to its value?

ugh92: yes it should and I am changing it so its `messageFromMessageKey(s"individualUserType.${IndividualUserType.Capacitor}")` since `messageFromMessageKey(s"individualUserType.agent.${IndividualUserType.Capacitor}")` doesn't refer to any message and will always give false positive

PR #294, opened 2020-03-20
[CPDP-1254] - Validation for lettings relief

modivilioglu:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/reliefdetails/ReliefDetailsController.scala
```diff
@@ -193,8 +194,16 @@ class ReliefDetailsController @Inject() (
                 draftReturn.copy(
                   reliefDetailsAnswers = Some(
                     answers.fold(
-                      _.copy(privateResidentsRelief = Some(AmountInPence.fromPounds(p))),
-                      _.copy(privateResidentsRelief = AmountInPence.fromPounds(p))
+                      _.copy(privateResidentsRelief = Some(AmountInPence.fromPounds(p)), lettingsRelief = None),
+                      c =>
+                        if (c.privateResidentsRelief === AmountInPence.fromPounds(p))
```
AndyHWChung: this check for equality should come outside before you do `draftReturn.copy` so we don't unecessarily unset the value of `lettingRelief` when the answer hasn't changed

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/finance/MoneyUtils.scala
```diff
@@ -54,6 +54,26 @@ object MoneyUtils {
         else Right(d)
       }
 
+  def validateLimit(key: String, exceedsLimit: BigDecimal => Boolean)(
+    s: String
```
AndyHWChung: Do you need `String` here? From the way you're using it it looks like you can always pass in a `BigDecimal`

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/finance/MoneyUtils.scala
```diff
@@ -54,6 +54,26 @@ object MoneyUtils {
...
+  ): Either[FormError, BigDecimal] =
+    Try(BigDecimal(cleanupAmountOfMoneyString(s))).toEither
+      .leftMap(_ => FormError(key, "error.invalid"))
+      .flatMap { d =>
+        if (exceedsLimit(d)) Left(FormError(key, "error.amountOverLimit"))
+        else Right(d)
+      }
+
+  def validateLimitLessThanOther(key: String, comparedOtherKey: String, exceedsLimit: BigDecimal => Boolean)(
```
AndyHWChung: this is doing the same thing as `validateLimit` - you could have a general method like this:
```scala
 def validateAmountOfMoney(key: String, isValid: BigDecimal => Boolean, errorMessageKey: => String)(value: BigDecimal) =
    if (isValid(d)) Right(d) else Left(FormError(key, errorMessageKey))
```
But that is so short that it might be easier not to have this method in the first place

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/reliefdetails/ReliefDetailsController.scala
```diff
@@ -205,51 +214,144 @@ class ReliefDetailsController @Inject() (
...
+      ),
+      routes.ReliefDetailsController.privateResidentsRelief()
+    )
+
+  private def lettingsPreBehaviour(
+    lettingsPostBehaviour: (
+      FillingOutReturn,
+      SingleDisposalDraftReturn,
+      ReliefDetailsAnswers,
+      AmountInPence
```
AndyHWChung: a lot of the stuff here is provided by `withFillingOutReturnAndReliefDetailsAnswers` already. You could just reuse that have a small method which allows you to get the private resident relief when you need it:
```scala
def withPrivateResidentRelief(answers: ReliefDetailsAnswers)(f: AmountInPence => Future[Result]): Future[Result] ???
```
You'll end up with a lot less code that way

AndyHWChung: then you're code would look something like this:
```
 withFillingOutReturnAndReliefDetailsAnswers(request) {
    case (_, fillingOutReturn, draftReturn, answers) =>
      withTaxYear(draftReturn){ taxYear =>
         ....
     }
}
```

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/reliefdetails/ReliefDetailsController.scala
modivilioglu: Done.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/finance/MoneyUtils.scala
modivilioglu: Key is used to mention the error message prefix. If there is a better way I may take that approach too.

AndyHWChung: I meant the `s` string here - but you might not even need this method if you take a look at my other comment

conf/messages
```diff
@@ -839,7 +839,8 @@ lettingsReliefValue.error.tooSmall=Amount of Letting Relief must be more than th
 lettingsReliefValue.error.tooLarge=Amount of Letting Relief must be 50,000,000,000.00 or less
 lettingsReliefValue.error.invalid=Amount of Letting Relief must be a number, like 108 or 2,940.54
 lettingsRelief.furtherInfoLink=Find out about Letting Relief and how to calculate it (opens in a new window)
-
+lettingsRelief.error.amountOverLimit=Amount of Letting Relief must be 40,000 or less
+lettingsRelief.error.amountOverPrivateResidenceRelief=Amount of Lettings Relief cannot be higher than private residence relief
```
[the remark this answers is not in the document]
modivilioglu: Done.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/reliefdetails/ReliefDetailsController.scala
modivilioglu: Yes, nice approach, done.

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/reliefdetails/ReliefDetailsControllerSpec.scala
```diff
@@ -439,6 +544,21 @@ class ReliefDetailsControllerSpec
             }
           }
         }
+
+        "the data is more than lettings relief limit" in {
+          test(
+            "lettingsRelief"      -> "0",
+            "lettingsReliefValue" -> "50000"
```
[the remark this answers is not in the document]
modivilioglu: Done.

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/reliefdetails/ReliefDetailsControllerSpec.scala
```diff
@@ -327,6 +355,60 @@ class ReliefDetailsControllerSpec
...
+              triageAnswers,
+              taxYear
+            )._1
+
+            inSequence {
+              mockAuthWithNoRetrievals()
+              mockGetSession(
+                session
+              )
+              val Some(FillingOutReturn(subscribedDetails, _, agentReferenceNumber, updatedDraftReturn)) =
```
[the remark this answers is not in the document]
modivilioglu: Yeah that's true, I am assuming it is already a Some where it might be a None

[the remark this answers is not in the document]
modivilioglu: It looks like I am already doing that in the "new overloaded sessionWithReliefDetailsAnswers". And then making sure in the storeDraftReturn that the values are send from the updatedSession. Am I missing something? :/

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/finance/MoneyUtils.scala
modivilioglu: Done.

conf/messages
```diff
@@ -839,7 +839,8 @@ lettingsReliefValue.error.tooSmall=Amount of Letting Relief must be more than th
 lettingsReliefValue.error.tooLarge=Amount of Letting Relief must be 50,000,000,000.00 or less
 lettingsReliefValue.error.invalid=Amount of Letting Relief must be a number, like 108 or 2,940.54
 lettingsRelief.furtherInfoLink=Find out about Letting Relief and how to calculate it (opens in a new window)
-
+lettingsRelief.error.amountOverLimit=Amount of Letting Relief must be 40,000 or less
```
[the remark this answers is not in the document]
modivilioglu: Done.

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/reliefdetails/ReliefDetailsControllerSpec.scala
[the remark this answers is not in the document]
AndyHWChung: I was talking about the `val Some(FillingOutReturn(...)) = updated.journeyStatus` being unsafe. Havking had a closer look, a safer way using what you have is just to use `val FillingOutReturn(...) = fillingOutReturn` - that is safer since there is no `Option` involved and `fillintOutReturn` is specifically set up as a `FillingOutReturn` rather than being typed as `JourneyStatus`. Being able to get a compile time error in case we get our types wrong is better than having a runtime MatchError.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/finance/MoneyUtils.scala
AndyHWChung: I'm think you can combine this function and the one above it to suit how you're using it in both your scenarios:
```scala
def validateValueIsLessThan(key: String, limitInclusive: AmountInPence, errorMessageKey: => String)(value: BigDecimal) =
  if(value <= limitInclusive) Right(d) else Left(FormError(key, errorMessageKey))
```
Then you can use it for both your scenarios:
```scala
    .flatMap(value =>
        MoneyUtils.validateValueIsLessThan(
          lettingsReliefLimit.inPounds(),
          "lettingsRelief",
          ,
          "error.lettingsReliefLimit"
        )(value)
      )
      .flatMap { value =>
        MoneyUtils.validateValueIsLessThan(
          "lettingsRelief",
          privateResidencyRelief.inPounds()
        )(
          value
        )
```
 The functions basically do the same thing but slightly differently

modivilioglu: Oh, ok, done.

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/reliefdetails/ReliefDetailsControllerSpec.scala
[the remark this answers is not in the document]
modivilioglu: Ohhhh yes right, I think I've done it this time. So rather than, getting the fillingOutReturn from the session as Option[JourneyStatus], we can directly get the FillingOutReturn from the sessionWithReliefDetailsAnswers. Done.

AndyHWChung:
LGTM

PR #297, opened 2020-03-20
cpdp-1348: submission error page, sent draft returns deleted

AndyHWChung:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/services/returns/ReturnsService.scala
```diff
@@ -94,19 +99,70 @@ class ReturnsServiceImpl @Inject() (connector: ReturnsConnector, auditService: A
...
+                         .parseJSON[GetDraftReturnResponse]()
+                         .leftMap(Error(_))
+                         .map(_.draftReturns)
+                     )
+      (sentDraftReturns, unsentDraftReturns) = draftReturns.partition(hasBeenSent(sentReturns))
+      _ <- if (sentDraftReturns.nonEmpty) EitherT.liftF[Future, Error, Unit](deleteSentDraftReturns(sentDraftReturns))
+          else EitherT.rightT[Future, Error](())
+    } yield unsentDraftReturns
+
+  private def deleteSentDraftReturns(sentDraftReturns: List[DraftReturn])(implicit hc: HeaderCarrier): Future[Unit] = {
```
AndyHWChung: we don't care if deletes are unsuccessful when we try to delete sent returns wince we are always yielding the unsentDraftReturns in the for-comprehension so the user will not see the sent draft returns anyway

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/CheckAllAnswersAndSubmitController.scala
```diff
@@ -72,33 +73,63 @@ class CheckAllAnswersAndSubmitController @Inject() (
           )
         val result =
           for {
-            response <- returnsService.submitReturn(submittedReturnRequest)
+            response <- returnsService
+                         .submitReturn(submittedReturnRequest)
+                         .map(Right(_))
+                         .leftFlatMap[Either[Error, SubmitReturnResponse], Error](e => EitherT.pure(Left(e)))
```
mikail-khan: I don't see the point of the `leftFlatMap` here. What have you achieved?

AndyHWChung: The error is now on the right of the either of the EitherT so I can change the journey status just below it .in the for-comprehension. It's not pretty, but the type of this part of the for-comprehension is `EitherT[Future, Error, Either[Error, SubmitReturnResponse]]`

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/CheckAllAnswersAndSubmitController.scala
```diff
@@ -72,33 +73,63 @@ class CheckAllAnswersAndSubmitController @Inject() (
...
+                    _.copy(journeyStatus = Some(newJourneyStatus))
                   )
                 )
-          } yield ()
+          } yield response
 
         result.fold(
           { e =>
-            logger.warn("Error while try to submit return and udpate session", e)
+            logger.warn("Error while trying to update session", e)
```
mikail-khan: `result` could fail with failed submission, so this error message is not really reflecting that

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/CheckAllAnswersAndSubmitController.scala
mikail-khan: Yeah, it's not good. I see why you have done it - you're trying to capture two different possible errors from two different effects in one for-comprehension. The problem - I think - boils down to the error channel in EitherT not being covariant. Because it is not and the fact that Error is final you cannot build a coproduct to reflect the different type of errors in the service e.g SessionUpdateError, SubmissionError, etc etc...

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/services/returns/ReturnsService.scala
AndyHWChung: The way the for comprehension works that submission errors always end up in the right of the EitherT in the Either. Do the type of the EitherT is `EitherT[Future, Error Either[Error, SubmitReturnResponse]]`

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/CheckAllAnswersAndSubmitController.scala
AndyHWChung: In this case it's more that you can't chain on the error type in the for comprehension with EitherT since flatMap always maps on the right type. That's why it looks like that - I wanted to be able to update the session on the for comprehension in the submit error case.

AndyHWChung: So even if I was able to build a coproduct it still won't allow me to map on the left in the EitherT in the for comprehension

mikail-khan: Do you think by changing the success type of `returnsService.submitReturn` to return the correct type you need i.e SubmitRetunFailed or JustSubmittedReturn and then case matching would be clearer?

AndyHWChung:
Whoops, I resolved the conversation by mistake

AndyHWChung:
I meant to write I'll do something with the types then to make it a bit clearer. I think I'll do it at the controller level though rather than the service level since this behaviour is specific to this controller

mikail-khan:
> I meant to write I'll do something with the types then to make it a bit clearer. I think I'll do it at the controller level though rather than the service level since this behaviour is specific to this controller

Yes but readability is important across all layers. Just to drag out a final observation: do you think that the real type of the submitReturn method (service layer) should be something like `EitherT[Future, Nothing, SomeTypeThatCapturesAllInformation]` - I think that is what we are really saying. Then the type system can do the heavy work of representing all cases we need to handle rather than coding them explicitly which is what you have done.

AndyHWChung:
I just pushed a commit. I think in general it's more flexible to leave the option open in the service layer to handle errors in a way in which the caller wants to handle it. If for instance we were to use the same method again elsewhere, we might not want to handle the errors in such a way in that scenario and we might be happier with an `Error` on the left of the `EitherT`. The way I changed it to now makes it a lot clearer now I think.

mikail-khan:
> I just pushed a commit. I think in general it's more flexible to leave the option open in the service layer to handle errors in a way in which the caller wants to handle it. If for instance we were to use the same method again elsewhere, we might not want to handle the errors in such a way in that scenario and we might be happier with an `Error` on the left of the `EitherT`. The way I changed it to now makes it a lot clearer now I think.

Let's go with this for now.

PR #304, opened 2020-03-22
shorten class names to avoid unucessary recompilation

AndyHWChung:


AndyHWChung:
all I've done is to change the name of a couple of classes

mikail-khan:
Let's give it a go - but I doubt this is the only reason why we have long compile times. Scalacheck Shapeless will still eat a lot of compilation cycles.

AndyHWChung:
Yeah, it's not going to reduce the compilation time of the generators, but it will reduce the number of times that file is compiled which helps.

PR #311, opened 2020-03-23
[CPDP-1254] - Additional behaviour when residentsrelief set to No

modivilioglu:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/reliefdetails/ReliefDetailsController.scala
```diff
@@ -333,10 +333,10 @@ class ReliefDetailsController @Inject() (
           page = otherReliefsPage(_, _)
         )(
           requiredPreviousAnswer = _.fold(
-            _.lettingsRelief,
-            c => Some(c.lettingsRelief)
+            _.privateResidentsRelief,
```
AndyHWChung: if PRR was greater than zero, then this should still be `lettingsRelief`

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/reliefdetails/ReliefDetailsControllerSpec.scala
```diff
@@ -182,7 +182,13 @@ class ReliefDetailsControllerSpec
       }
 
     }
+    "handling check your answers" must {
```
AndyHWChung: there should already be a section in the spec for the check your answers page. Within that section is already an area where it's testing when we should be redirected to the other reliefs page (see L1188) - these tests belong there. I'd take a look at those tests to see how it's being done. Also, you don't need to have an form data in the body for check your answers

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/reliefdetails/ReliefDetailsControllerSpec.scala
```diff
@@ -182,7 +182,13 @@ class ReliefDetailsControllerSpec
...
+      def performAction(data: Seq[(String, String)]): Future[Result] =
+        controller.checkYourAnswers()(FakeRequest().withFormUrlEncodedBody(data: _*))
 
+      behave like residentsReliefSubmittedYesBehaviour(performAction, AmountInPence(500))
```
AndyHWChung: you're only using these functions once - there's not much point in having them. If you want to see what the test is actually doing you have to go to the function which adds a level of indirection to the code

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/reliefdetails/ReliefDetailsControllerSpec.scala
```diff
@@ -1167,7 +1173,8 @@ class ReliefDetailsControllerSpec
             mockAuthWithNoRetrievals()
             mockGetSession(
               sessionWithReliefDetailsAnswers(
-                allQuestionsAnswered.copy(lettingsRelief = None)
+                allQuestionsAnswered
+                  .copy(privateResidentsRelief = Some(AmountInPence(20)), lettingsRelief = None, otherReliefs = None)
```
AndyHWChung: it'd be clearer if the spec was updated to say something like "the user has not answered that question and the amount of private residents relief is greater them zero"

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/reliefdetails/ReliefDetailsControllerSpec.scala
```diff
@@ -1355,7 +1362,50 @@ class ReliefDetailsControllerSpec
...
+        val session = sessionWithReliefDetailsAnswers(
+          fillingOutReturn,
+          singleDisposalDraftReturn,
+          Some(incompleteAnswers),
+          None,
+          disposalDate,
+          triageAnswers,
+          taxYear
+        )._1
+        val data = Seq(
```
AndyHWChung: you don't need `data` here - you're doing this test of the display check your answers endpiont which doesn't take any form data. The data necessary for this spec is controlled via what's in session.

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/reliefdetails/ReliefDetailsControllerSpec.scala
```diff
@@ -1355,7 +1362,50 @@ class ReliefDetailsControllerSpec
         }
 
         checkIsRedirect(
-          performAction(),
+          performAction(Seq("privateResidentsRelief" -> "1")),
+          routes.ReliefDetailsController.otherReliefsSubmit()
```
AndyHWChung: this is weird - why would it be redirecting to the submit endpoint? Is this test actually passing?

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/reliefdetails/ReliefDetailsController.scala
modivilioglu: ooooh yess. good point.

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/reliefdetails/ReliefDetailsControllerSpec.scala
modivilioglu: The reason it is passing should be that submit also redirects to residents relief page in case that does not exist. The same is done on the display endpoint too. This is the same behaviour when it was the lettingsRelief. It makes more sense to make it otherReliefs() rather than otherReliefsSubmit() though as per the thing we are trying to mimic.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/reliefdetails/ReliefDetailsController.scala
modivilioglu: Hmmmm... I might change the signature though, as now this time, the Call will be dependant on the required parameter.

AndyHWChung: I think you probably to change it to something like `ReliefDetailsAnswers => Boolean` to avoid wart remover complaining about products and having something like this:
```scala
 hasRequiredPreviousAnswer = { answers =>
   if(privateResidentsReliefIsZero) answers.fold(_.privateResidentsRelief, ...).isDefined  
   else answers.fold(_.lettingsRelief, ...).isDefined
  },
redirectIfNoPreviousAnswers = { answers =>
     if(privateResidentsReliefIsZero) privateResidentsReliefCall  
   else lettingReliefCall
}

```

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/reliefdetails/ReliefDetailsControllerSpec.scala
modivilioglu: Absolutely. Done.

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/reliefdetails/ReliefDetailsControllerSpec.scala
modivilioglu: Done, also I've noticed that the case residentsReliefSubmittedYesBehaviour was already being handled in the code. So removed it.

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/reliefdetails/ReliefDetailsControllerSpec.scala
modivilioglu: Yes. I think I was in a rush yesterday to get this done to move on with 1252, that is why I missed that.. Done.

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/reliefdetails/ReliefDetailsControllerSpec.scala
modivilioglu: looks like I've forgotten that while I was in a rush. That was a silly mistake on my side. Done.

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/reliefdetails/ReliefDetailsControllerSpec.scala
modivilioglu: Done.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/reliefdetails/ReliefDetailsController.scala
modivilioglu: Why not to ReliefDetailsAnswers => Call? Thanks

modivilioglu: Done, there seems to be many scenarios with the new approach. Maybe we might need to discuss it. But done.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/reliefdetails/ReliefDetailsController.scala
```diff
@@ -92,16 +92,16 @@ class ReliefDetailsController @Inject() (
     page: (Form[A], Call) => P
   )(
     requiredPreviousAnswer: ReliefDetailsAnswers => Option[R],
-    redirectToIfNoRequiredPreviousAnswer: Call
+    redirectToIfNoRequiredPreviousAnswer: ReliefDetailsAnswers => Call
```
AndyHWChung: you might find it easier if you change 
```scala
    requiredPreviousAnswer: ReliefDetailsAnswers => Option[R]
```
to 
```scala
    hasRequiredPreviousAnswer: ReliefDetailsAnswers => Boolean
```
The value of the previous answer is never used - we only care about whether or not the value is there or not

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/reliefdetails/ReliefDetailsControllerSpec.scala
AndyHWChung: is `data` still being used somewhere?

modivilioglu: No, good spot, Deleted.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/reliefdetails/ReliefDetailsController.scala
modivilioglu: Done

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/reliefdetails/ReliefDetailsController.scala
```diff
@@ -381,6 +369,54 @@ class ReliefDetailsController @Inject() (
     }
   }
 
+  private def hasRequiredPreviousAnswerForOtherReliefs(answers: ReliefDetailsAnswers): Boolean =
+    answers match {
```
AndyHWChung: I think this is more easily read like this:
```scala
val privateResidentsRelief = answers.fold(_.privateResidentsRelief, c => Some(c.private...))
val lettingRelief =answers.fold(_.lettingRelief, c => Some(c.lettingRelief)

privateResidentRelief.exists(_.isZero) || (
  privateResidentsRelief.exists(_  > AmountInPence.zero) && lettingRelief.isDefined
)  
```
what do you think?

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/reliefdetails/ReliefDetailsController.scala
```diff
@@ -381,6 +369,54 @@ class ReliefDetailsController @Inject() (
...
+        routes.ReliefDetailsController.privateResidentsRelief()
+      case IncompleteReliefDetailsAnswers(Some(AmountInPence(_)), lettingsRelief, _) =>
+        routes.ReliefDetailsController.lettingsRelief()
+      case CompleteReliefDetailsAnswers(AmountInPence(0), lettingsRelief, _) if !lettingsRelief.isZero =>
+        routes.ReliefDetailsController.privateResidentsRelief()
+      case CompleteReliefDetailsAnswers(AmountInPence(_), _, _) => routes.ReliefDetailsController.lettingsRelief()
+    }
+
+  private def hasRequiredPreviousAnswerForLettingsReliefs(answers: ReliefDetailsAnswers): Boolean =
+    answers match {
```
AndyHWChung: you probably don't need all these cases. The previous question to letting relief is always private residents relief, so this becomes simply `answers.fold(_.privateResidentsRelief, c => Some(c.privateResidentsRelief)).isDefined`

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/reliefdetails/ReliefDetailsController.scala
```diff
@@ -381,6 +369,54 @@ class ReliefDetailsController @Inject() (
...
+      case IncompleteReliefDetailsAnswers(Some(AmountInPence(0)), None, _)                   => false
+      case IncompleteReliefDetailsAnswers(Some(AmountInPence(0)), Some(AmountInPence(0)), _) => false
+      case IncompleteReliefDetailsAnswers(Some(AmountInPence(0)), Some(_), _)                => false
+      case IncompleteReliefDetailsAnswers(Some(AmountInPence(_)), _, _)                      => true
+      case CompleteReliefDetailsAnswers(AmountInPence(0), AmountInPence(0), _)               => false
+      case CompleteReliefDetailsAnswers(_, _, _)                                             => true
+    }
+
+  private def getRedirectForLettingsReliefs(answers: ReliefDetailsAnswers): Call =
+    answers match {
```
AndyHWChung: this should always point to private residents relief

modivilioglu: Done.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/reliefdetails/ReliefDetailsController.scala
modivilioglu: Done.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/reliefdetails/ReliefDetailsController.scala
modivilioglu: Done.

AndyHWChung:
LGTM

PR #313, opened 2020-03-24
cpdp-1400: single disposal non-calculated return now completeable

AndyHWChung:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/CheckAllAnswersAndSubmitController.scala
```diff
@@ -30,7 +30,7 @@ import uk.gov.hmrc.cgtpropertydisposalsfrontend.controllers.accounts.homepage
 import uk.gov.hmrc.cgtpropertydisposalsfrontend.controllers.returns.acquisitiondetails.RebasingEligibilityUtil
 import uk.gov.hmrc.cgtpropertydisposalsfrontend.models.JourneyStatus.{FillingOutReturn, JustSubmittedReturn}
 import uk.gov.hmrc.cgtpropertydisposalsfrontend.models.SessionData
-import uk.gov.hmrc.cgtpropertydisposalsfrontend.models.returns.{CompleteReturn, MultipleDisposalsDraftReturn, SingleDisposalDraftReturn, SubmitReturnRequest}
+import uk.gov.hmrc.cgtpropertydisposalsfrontend.models.returns.{CompleteSingleDisposalReturn, MultipleDisposalsDraftReturn, SingleDisposalDraftReturn, SubmitReturnRequest}
```
mikail-khan: Did you manage to figure out the limit before zinc triggers re-compiles? Just thinking that identifier might be too long.

AndyHWChung: I didn't in the end - I think the limit is on the actual class name excluding the packages in the path. I'll double check but I don't think these names take us over the limit

AndyHWChung: just checked  - we're still good on the limit

PR #317, opened 2020-03-25
CPDP-1423: Disposal date error message fix

rambabu-posa:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/address/PropertyDetailsController.scala
```diff
@@ -726,6 +730,11 @@ class PropertyDetailsController @Inject() (
         _ => routes.PropertyDetailsController.checkYourAnswers()
       )
 
+  private def getCompletionDate(answers: MultipleDisposalsTriageAnswers): CompletionDate =
+    answers
+      .fold(_.completionDate, c => Some(c.completionDate))
+      .getOrElse(CompletionDate(LocalDateUtils.today()))
```
AndyHWChung: we shouldn't default to today - we should redirect to the task list if we can't find a completion date in the triage answers

AndyHWChung: in your pattern match for the tax year you could possible add in a check for the completion date as well at the same time:
```scala
 m.triageAnswers.fold(i => i.taxYear -> i.completionDate, c => Some(c.taxYear) -> Some(c.completionDate))) match {
  case (Some(taxYear), Some(completionDate)) => ...
   case _ => Redirect(taskList)
```

rambabu-posa: Yes I liked this approach

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/address/PropertyDetailsController.scala
```diff
@@ -384,19 +384,21 @@ class PropertyDetailsController @Inject() (
         r.draftReturn match {
           case _: SingleDisposalDraftReturn => Redirect(routes.PropertyDetailsController.checkYourAnswers())
           case m: MultipleDisposalsDraftReturn =>
-            m.triageAnswers.fold(_.taxYear, c => Some(c.taxYear)) match {
-              case Some(taxYear) =>
+            m.triageAnswers.fold(
+              i => i.taxYear       -> i.completionDate,
```
AndyHWChung: I think we're just missing unit tests to say what happens when the completion date isn't there now

AndyHWChung:
LGTM

PR #320, opened 2020-03-25
Acquisition price cya rename

sam-miele:
Made a mistake with the last one. This question was meant for the acquisition price, not the date


AndyHWChung:
LGTM

PR #323, opened 2020-03-26
cpdp-1400: removed return to summary link on big cya page

AndyHWChung:


AndyHWChung:
missing a couple of links - will put them in now

PR #325, opened 2020-03-26
[CPDP-1254] - Do not show link to lettings relief when residents reli…

modivilioglu:
…ef is selected as no


test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/reliefdetails/ReliefDetailsControllerSpec.scala
```diff
@@ -1379,7 +1379,29 @@ class ReliefDetailsControllerSpec
             )
           }
         }
+        "the user has set residential relief to no" in {
```
AndyHWChung: probably don't need this test - it should be covered by the `forall` in the test above. Also, in this test there would have been no need to do a `forall` since you're populating the  answers with a specific one each time rather than using the generated one

modivilioglu: I agree with forall, but looks like forall does not test the AmountInPence(0) case. Thanks

AndyHWChung:
LGTM

PR #332, opened 2020-03-27
1374-fix-check-your-answer-view

sam-miele:


test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/exemptionandlosses/ExemptionAndLossesControllerSpec.scala
```diff
@@ -1483,6 +1484,18 @@ object ExemptionAndLossesControllerSpec extends Matchers {
       )
     }
 
+    if (completeExemptionAndLossesAnswers.annualExemptAmount.isPositive) {
```
AndyHWChung: do we not want to show this question even if the user input zero?

AndyHWChung: wait, this is a spec - I meant isn't this question and answer shown even if the user enters zero?

AndyHWChung: the reason why the ones above have the positivity-guard around them is that they're set to zero is that the user is not allowed to enter zero values if they select yes to those questions. We use the value zero to signify the user selected no for those questions. The annual exempt question is different since it's not a yes-no questions and the user can enter zero if they want.

AndyHWChung:
LGTM

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/exemptionandlosses/ExemptionAndLossesControllerSpec.scala
sam-miele: I checked. Works fine with 0. Was a hangover from trying to debug why some of the tests were failing. Removed now 😄

PR #335, opened 2020-03-28
Cpdp 1250 field dependencies

AndyHWChung:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/disposaldetails/DisposalDetailsController.scala
```diff
@@ -254,16 +269,26 @@ class DisposalDetailsController @Inject() (
...
-                  d.fold(
-                    _.copy(disposalPrice = Some(fromPounds(price))),
-                    _.copy(disposalPrice = fromPounds(price))
-                  )
-              },
-              nextPage = _.fold(
-                _ => controllers.returns.disposaldetails.routes.DisposalDetailsController.whatWereDisposalFees(),
-                _ => controllers.returns.disposaldetails.routes.DisposalDetailsController.checkYourAnswers()
-              )
+                case (price, answers, draftReturn) =>
```
mikail-khan: you can remove this case keyword - it's not required - only required if you have multiple different lambdas you wanted to match against.

AndyHWChung: I've gone and removed a load of cases where we didn't need them

PR #344, opened 2020-03-30
CPDP-1398: Fixes for Returns tasklist statuses unit tests

rambabu-posa:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/single_disposal_task_list.scala.html
```diff
@@ -75,11 +76,11 @@
     }
 }
 
-
 @reliefDetailsState = @{
-    (disposalDetailsState, acquistionDetailsState,initialGainOrLossState) match {
```
AndyHWChung: is there a unit test to test this new behaviour? I'd expect to see a minimum of one test testing that if the other three sectinos are compelte,  then the relief section is not  `CannotStart` if the initial gain or loss section is not complete if the initial gain or loss section is not required and another testing that it is `CannotStart` if the initial gain or loss section is not complete and that section is required

rambabu-posa: Yes there are two tests to check reliefDetailsAnswers status for CannotStart

AndyHWChung: the test coverage there does not look complete. One test is testing what happens if the disposal details section is not complete and the other test is tseting if the acquisition details section is not complete. There are missing tests for the cannot start state for when:
- the property details section is not complete
- the initial gain or loss section is not complete and the initial gain or loss section is required
For that last one, you will need to modify the answers in the return to be for a non uk resident disposing of a residential property with a certain acquisition date

AndyHWChung:
LGTM

PR #345, opened 2020-03-30
Cpdp 1414 draft return saved content updates

AndyHWChung:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/ReturnSavedController.scala
```diff
@@ -33,20 +33,20 @@ import uk.gov.hmrc.play.bootstrap.controller.FrontendController
 import scala.concurrent.{ExecutionContext, Future}
 
 @Singleton
-class ConfirmDraftReturnController @Inject() (
+class ReturnSavedController @Inject() (
```
mikail-khan: what was the reason for the name change?

AndyHWChung: the first one sounded funny - what's actually happening is we're telling the user that the draft return has been saved. "Confirming" a draft return doesn't quite make sense

mikail-khan: could you rename this to `DraftReturnSavedController` pls?

conf/returns.routes
```diff
@@ -21,4 +21,4 @@ GET   /pay-charge-in-return              uk.gov.hmrc.cgtpropertydisposalsfronten
 
 GET   /check-if-return-was-sent          uk.gov.hmrc.cgtpropertydisposalsfrontend.controllers.returns.CheckAllAnswersAndSubmitController.submissionError()
 POST  /check-if-return-was-sent          uk.gov.hmrc.cgtpropertydisposalsfrontend.controllers.returns.CheckAllAnswersAndSubmitController.submissionErrorSubmit()
-GET   /return-draft-confirmation         uk.gov.hmrc.cgtpropertydisposalsfrontend.controllers.returns.ConfirmDraftReturnController.confirmDraftReturn()
+GET   /exit/return-saved                 uk.gov.hmrc.cgtpropertydisposalsfrontend.controllers.returns.ReturnSavedController.draftReturnSaved()
```
mikail-khan: change endpoint to `/exit/draft-return-saved`

AndyHWChung: the url came from Jean - I'll leave it like that unless Jean wants it to be changed

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/ReturnSavedController.scala
AndyHWChung: done

PR #347, opened 2020-03-30
[CPDP-1252] - Test bug fix

modivilioglu:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/ytdliability/estimated_income.scala.html
```diff
@@ -69,7 +69,6 @@
 
         @details(messages(s"$key.details.summary"),
             Html(s"""|
-            |<p>${messages(s"$key.details.p1", taxYearStartYear, taxYearEndYear)}</p>
```
AndyHWChung: Do we not need p1 anymore? We can delete this from the messages file then to clean it up if we don't. Also then it'll look nicer if p2 became p1 in the key if you delete p1.

modivilioglu: Sure.

modivilioglu: Done.

AndyHWChung:
LGTM

PR #351, opened 2020-03-31
CPDP-1372: Disposal details custom content changes

rambabu-posa:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/disposaldetails/how_much_did_you_own.scala.html
```diff
@@ -33,10 +34,15 @@
...
+@userKey = @{if (isAgent) ".agent" else if (isATrust) ".trust" else ""}
+@userErrorPrefix =  @{if (isAgent) "agent." else if (isATrust) "trust." else ""}
+
 @key = @{ "shareOfProperty" }
 @percentageShareKey = @{"percentageShare"}
-@title = @{messages(s"$key.title")}
+@title = @{messages(s"$key$userKey.title")}
 @hasErrors = @{form.hasErrors || form.hasGlobalErrors}
 @options = @{List(
   RadioOption(
```
nubz, line 48: the conditional content has a currency input that is missing a customErrorKey on it - the errors on that field are not customised for user types otherwise

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/disposaldetails/how_much_did_you_own.scala.html
```diff
@@ -91,9 +97,10 @@
         fieldId = key,
         label = pageHeading(title),
         selected = selectedIndex,
-          error = form.error(key),
+        error = form.error(key),
         hasErrors = hasErrors,
-        options = options
+        options = options,
+        customErrorKey = form.error(key).map(e => s"$userErrorPrefix${e.message}")
```
nubz, line 103: this customErrorKey will not work - it should include the key and then the userKey like `customErrorKey = Some(s"$key$userKey")`

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/disposaldetails/how_much_did_you_own.scala.html
```diff
@@ -33,10 +34,15 @@
   caption: uk.gov.hmrc.cgtpropertydisposalsfrontend.views.html.components.caption
 )
 
-@(form: Form[ShareOfProperty], backLink: Call)(implicit request: RequestWithSessionData[_], messages:Messages, appConfig: ViewConfig)
+@(form: Form[ShareOfProperty], backLink: Call, isATrust: Boolean)(implicit request: RequestWithSessionData[_], messages:Messages, appConfig: ViewConfig)
+
+@isAgent = @{ request.userType.contains(UserType.Agent) }
+@userKey = @{if (isAgent) ".agent" else if (isATrust) ".trust" else ""}
+@userErrorPrefix =  @{if (isAgent) "agent." else if (isATrust) "trust." else ""}
```
nubz, line 41: this is not needed in this template - we can just pass customErrorKeys to the components in here as they both accept error objects and then look for a customErrorKey - see the way components.radio_group_conditional and components.unit_input consume customErrorKey - other components might need this (when they pass in the entire error message key) but not the ones in here

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/disposaldetails/DisposalDetailsControllerSpec.scala
```diff
@@ -102,6 +109,60 @@ class DisposalDetailsControllerSpec
...
+  def sessionWithDisposalDetailsAnswersForIndividual(
+    answers: DisposalDetailsAnswers,
+    disposalMethod: DisposalMethod
+  ): (SessionData, FillingOutReturn, DraftSingleDisposalReturn) = {
+
+    val subscribedDetails = sample[SubscribedDetails].copy(
+      name = Right(sample[IndividualName])
+    )
+
+    sessionWithDisposalDetailsAnswers(Some(answers), disposalMethod, subscribedDetails)
```
nubz, line 151: I can't see how you are ensuring the userType in the session is UserType.Individual so you may end up with a sampled UserType such as UserType.Organisation or UserType.Agent in your session - we need to do more than just set the name to match our userType, this may mean passing a new param in to `sessionWithDisposalDetailsAnswers` such as `userType: UserType` and set it in this call to `UserType.Individual` then you can add in the necessary `userType = Some(userType)` to the session when copying extra vals in - best to look at other specs that handle this where we use methods like `setAgentReferenceNumber(userType)` because we want to control this also

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/disposaldetails/disposal_fees.scala.html
```diff
@@ -55,12 +61,13 @@
...
             labelAsHeading = true,
-            helpText = Some(Html(messages(s"$key.helpText"))),
+            helpText = Some(Html(messages(s"$key$userKey.helpText"))),
             autoComplete = AutoCompleteType.On,
             inputClass = Some("small"),
             aspect = "left",
             unit = "£",
-            assistiveText = messages("unit-input.assistive.pounds")
+            assistiveText = messages("unit-input.assistive.pounds"),
+            customErrorKey = form.error(key).map(e => s"$userErrorPrefix${e.message}")
```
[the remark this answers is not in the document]
AndyHWChung: if you haven't already done so, it'd b useful to manually check to see the error messages render correctly on the browser for all the pages you have worked on for all user types

rambabu-posa:
> there is quite a lot not working here - it may compile but messages are not being used because keys are not set correctly

all are working now

AndyHWChung:
LGTM

PR #352, opened 2020-03-31
Cpdp 1317

sam-miele:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/components/return_banner.scala.html
```diff
@@ -16,21 +16,33 @@
...
+@import uk.gov.hmrc.cgtpropertydisposalsfrontend.models.onboarding.SubscribedDetails
+@import uk.gov.hmrc.cgtpropertydisposalsfrontend.models.returns.ReturnSummary
 
 @this()
 
-@(heading: String, taxOwed: AmountInPence, reference: String)(implicit messages: Messages)
+@(heading: String, returnSummary: ReturnSummary, subscribedDetails: SubscribedDetails, userType: Option[UserType])(implicit messages: Messages)
+
+@taxOwed = @{returnSummary.mainReturnChargeAmount.withFloorZero}
+@name = @{subscribedDetails.name.fold(_.value, e => e.firstName + " " + e.lastName)}
```
AndyHWChung: not sure if you should use this or the `contactName` field - best to double check

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/confirmation_of_submission.scala.html
```diff
@@ -21,34 +21,62 @@
...
         exitSurveyLink: uk.gov.hmrc.cgtpropertydisposalsfrontend.views.html.components.exit_survey,
+        addressDisplay: uk.gov.hmrc.cgtpropertydisposalsfrontend.views.html.components.address_display,
         warning: uk.gov.hmrc.cgtpropertydisposalsfrontend.views.html.components.warning
 )
 
 @(justSubmittedReturn: JustSubmittedReturn)(implicit request: RequestWithSessionData[_], messages:Messages, viewConfig: ViewConfig)
 
 @key=@{"confirmationOfSubmission"}
 @title = @{ messages(s"$key.title") }
+@name = @{justSubmittedReturn.subscribedDetails.name.fold(_.value, e => e.firstName + " " + e.lastName)}
```
AndyHWChung: same here, might just want contact name

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/confirmation_of_submission.scala.html
```diff
@@ -21,34 +21,62 @@
...
     </strong>
   </span>
   <span class="heading-large" style="font-weight: normal;">
-   @messages(s"$key.returnReference")
-   <strong style="display:block; font-weight: bold;">@{{justSubmittedReturn.submissionResponse.formBundleId}}</strong>
+      <div id="ref-id">
+       @{if(amountOwed.equals(BigDecimal(0))) {
+               messages(s"$key.returnReference")
+           } else {
+                "Payment reference number"
```
AndyHWChung: this needs to be read from the messages file

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/confirmation_of_submission.scala.html
```diff
@@ -60,6 +88,47 @@ <h1 class="heading-xlarge">@title</h1>
    <a href="JavaScript: window.print();">@messages(s"$key.printPage")</a>
   </p>
 
+    <table class="borders payment-table">
+        <tbody id="return-details-table">
+
+
+        <tr class="charge">
+            <td> <strong class="bold"> Return sent to HMRC </strong> <br/> </td>
```
AndyHWChung: this should come from messages too

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/confirmation_of_submission.scala.html
```diff
@@ -60,6 +88,47 @@ <h1 class="heading-xlarge">@title</h1>
...
+            <td>
+                @LocalDateUtils.govDisplayFormat(justSubmittedReturn.submissionResponse.processingDate.toLocalDate)
+            </td>
+        </tr>
+
+        @{if(!amountOwed.equals(BigDecimal(0))) {
+        <tr class="charge">
+            <td> <strong class="bold">Return reference number</strong> <br/> </td>
```
AndyHWChung: and this one

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/confirmation_of_submission.scala.html
```diff
@@ -60,6 +88,47 @@ <h1 class="heading-xlarge">@title</h1>
...
+            <td> {justSubmittedReturn.submissionResponse.formBundleId} </td>
+        </tr>
+        }}
+
+
+        <tr class="charge">
+            <td> <strong class="bold">Property address</strong> <br/> </td>
```
AndyHWChung: and also this one

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/confirmation_of_submission.scala.html
```diff
@@ -60,6 +88,47 @@ <h1 class="heading-xlarge">@title</h1>
...
+                    case c: CompleteMultipleDisposalsReturn => addressDisplay(c.examplePropertyDetailsAnswers.address, Line)
+                    case c: CompleteSingleDisposalReturn => addressDisplay(c.propertyAddress, Line)
+                }}
+            </td>
+        </tr>
+
+        @{if(!amountOwed.equals(BigDecimal(0))) {
+            {justSubmittedReturn.submissionResponse.charge.map { charge =>
+                <tr class = "charge">
+                    <td> <strong class="bold">Tax due by</strong> <br/>
```
AndyHWChung: aaand this one

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/components/return_banner.scala.html
```diff
@@ -16,21 +16,33 @@
...
+@reference = @{if(taxOwed.isPositive) subscribedDetails.cgtReference.value else returnSummary.submissionId}
 
 <div class="govuk-box-highlight govuk-box-highlight--status">
-  <h1 class="heading-xlarge">@heading</h1>
+    @{userType match {
+        case Some(UserType.Organisation) => "Trust: " + name
```
AndyHWChung: "Trust" and "Client" should really be read from the messages as well like:
```
message.trust.key=Trust: {0}
message.agent.key=Client: {0}
```
basically there should be no hard coded strings in the templates that get displayed to the user - when we do a translation of the service, it's the message file which will get translated so everything needs to be in there

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/CheckAllAnswersAndSubmitControllerSpec.scala
```diff
@@ -485,22 +491,132 @@ class CheckAllAnswersAndSubmitControllerSpec
...
+        val noTaxDueRefLine             = "Return reference number form bundle id"
+        val taxDueRefLine               = "Payment reference number charge ref"
+        val submissionLine              = "Return sent to HMRC 2 February 2020"
+        val addressLine                 = "Property address 123 fake street, abc123"
+        val returnReferenceWithBundleId = "Return reference number form bundle id"
+        val taxDueDateLine              = "Tax due by 1 January 2020"
+
+        sealed case class TestScenario(
+          description: String,
+          userType: Option[UserType],
```
AndyHWChung: you're never using `None` so it might be worth just having this as non-optional

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/CheckAllAnswersAndSubmitControllerSpec.scala
```diff
@@ -485,22 +491,132 @@ class CheckAllAnswersAndSubmitControllerSpec
...
+                processingDate = LocalDateTime.of(processingDate, LocalTime.of(1, 1)),
+                charge = Some(
+                  sample[ReturnCharge].copy(
+                    amount          = taxOwed,
+                    chargeReference = "charge ref",
+                    dueDate         = dueDate
+                  )
+                )
+              )
+              val subscribe = sample[SubscribedDetails].copy(name = Right(IndividualName("John", "Doe")))
```
AndyHWChung: if the user type is trust, it'd be more consitent to have the name as something like `Left(sample[TrustName])`

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/CheckAllAnswersAndSubmitControllerSpec.scala
```diff
@@ -485,22 +491,132 @@ class CheckAllAnswersAndSubmitControllerSpec
...
+              val justSubmittedReturn = sample[JustSubmittedReturn].copy(
+                submissionResponse = returnResponse,
+                subscribedDetails  = subscribe,
+                completeReturn     = sample[CompleteSingleDisposalReturn].copy(propertyAddress = address)
```
AndyHWChung: if we have a user type of agent can we also populate the agentReference number with some value here for consistency? So something like 
```scala
 .copy(
    ...
    agentReferenceNumber = if(userType === UserType.Agent) Some(sample[AgentReferenceNumber]) else None
)
```

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/components/return_banner.scala.html
nubz: in this context it is correct to use `name` - it's to show agents and trust the name of the account (not the contact name) the return was posted against

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/confirmation_of_submission.scala.html
nubz: same here, `name` is to communicate the account name not the contact - there's a method on IndividualName  `e.makeSingleName()` so you don't need to concatenate first and last names each time

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/components/return_banner.scala.html
AndyHWChung: this still need a similar treatment that you've applied to the other template:
```scala
 @{request.userType match {
         case Some(UserType.Organisation) => messages(s"$key.organisationNameLine", name)
         case Some(UserType.Agent) => messages(s"$key.agentNameLine", name)
         case _ => name
     }}
```

AndyHWChung: that's if we need a name for an individual - can double check if we need it here for that case?

AndyHWChung:
LGTM

PR #384, opened 2020-04-03
Cpdp 1161 upload mandatory evidence

AndyHWChung:


conf/messages
```diff
@@ -1451,8 +1451,8 @@ calculator.total.amountDueAtHigherRate=Amount of tax due at the higher rate
 calculator.total.taxOwed=Capital Gains Tax due for this return
 calculator.noTaxToPay=No tax is due for this return
 
-mandatoryEvidence.title=Add a document showing how you worked out the tax
-mandatoryEvidence.helpText=Upload one or more documents (maximum file size 5MB) showing how you worked out the Capital Gains Tax. For example, a spreadsheet or PDF.
+mandatoryEvidence.title=Upload a document showing how the Capital gains Tax due was worked out
+mandatoryEvidence.helpText=Files can be up to 5MB. We accept these file types XLS, XLSX, ODS, PDF, DOC, DOCX, ODT, TXT, PNG and JPG, Any files you upload to a draft return will be saved for 7 days.
```
mikail-khan, line 1455: I thought the messages text would be the same in both scenarios?

AndyHWChung: this one looks like the content is more tailored to talk about a file showing how they calculated their tax. The mandatory evidence is required when the user does not use a figure we calculate.

AndyHWChung: I think I lifted these from the ticket anyway, but I'll double check

mikail-khan: Yeah it doesn't look right. Because this page should have any context other than uploading a file

AndyHWChung: not sure I get what you mean

PR #390, opened 2020-04-03
CPDP-1271: add boundary separator

mikail-khan:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/connectors/UpscanConnector.scala
```diff
@@ -279,6 +282,21 @@ class UpscanConnectorImpl @Inject() (
     implicit hc: HeaderCarrier
   ): EitherT[Future, Error, Unit] = {
 
+    implicit val actorSystem       = ActorSystem()
+    implicit val actorMaterializer = ActorMaterializer()
```
mikail-khan, line 286: added this locally here because injecting it into the class broke 70 test suites :man_shrugging:  - if this works then can inject and fix tests.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/connectors/UpscanConnector.scala
```diff
@@ -279,6 +282,21 @@ class UpscanConnectorImpl @Inject() (
     implicit hc: HeaderCarrier
   ): EitherT[Future, Error, Unit] = {
 
+    implicit val actorSystem       = ActorSystem()
```
AndyHWChung, line 285: won't this start a whole new `ActorSystem`? If you need a materializer I think you might be able to inject one in to the class. If you can't you should be able to inject in the actor system so you don't start a whole new one every time this method is called

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/connectors/UpscanConnector.scala
```diff
@@ -279,6 +282,21 @@ class UpscanConnectorImpl @Inject() (
...
+    implicit val actorMaterializer = ActorMaterializer()
+
+    implicit val multipartBodyWriter: BodyWritable[Source[MultipartFormData.Part[Source[ByteString, _]], _]] = {
+      val boundary    = Multipart.randomBoundary()
+      val contentType = s"multipart/form-data; boundary=$boundary"
+      BodyWritable(
+        body => {
+          val byteString = Multipart.transform(body, boundary).runFold(ByteString.empty)(_ ++ _)
+          InMemoryBody(Await.result(byteString, Duration(90, SECONDS)))
```
AndyHWChung, line 294: can't we map into the `Future` here instead of awaiting?

PR #400, opened 2020-04-04
Cpdp 1463 view return handling attachment display

nubz:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/returns/CompleteReturn.scala
```diff
@@ -43,7 +43,8 @@ object CompleteReturn {
     examplePropertyDetailsAnswers: CompleteExamplePropertyDetailsAnswers,
     exemptionAndLossesAnswers: CompleteExemptionAndLossesAnswers,
     yearToDateLiabilityAnswers: CompleteNonCalculatedYTDAnswers,
-    uploadSupportingDocumentAnswers: CompleteUploadSupportingEvidenceAnswers
+    uploadSupportingDocumentAnswers: CompleteUploadSupportingEvidenceAnswers,
+    hasAttachments: Boolean = true
```
AndyHWChung, line 47: can we avoid a default arguments? It's generally not safe to have them in case we forget to set it and it gets the wrong value

nubz: no problem

AndyHWChung:
LGTM

PR #421, opened 2020-04-09
CPDP-1497: ExemptionAndLosses error messages fix 

rambabu-posa:


test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/exemptionandlosses/ExemptionAndLossesControllerSpec.scala
```diff
@@ -1424,6 +1425,12 @@ class ExemptionAndLossesControllerSpec
           errorArgs: _*
         )
         doc.title() should startWith("Error:")
+
+        val topErrorMsg     = doc.select(s"""a[href="#$key"]""").text()
+        val inlineErrorMsg  = doc.select(s"""span[id="$key-inline-error"]""").text()
+        val topErrorMessage = if (topErrorMsg.nonEmpty) "Error: " + topErrorMsg else topErrorMsg
+
+        topErrorMessage shouldEqual inlineErrorMsg
```
AndyHWChung, line 1433: could we rename references to `topError` to `errorSummary` so it's clearer. Also, if we're testing a form error, the error summary should always have something in it, so I'm not sure if testing for an empty error summary is appropriate

AndyHWChung: the inline error message wouldn't include the "Error: " prefix would it?

rambabu-posa: When I inspect inline error message its something like this

```
<span class="error-message" id="annualExemptAmount-inline-error">
  <span class="visuallyhidden">Error: </span>
  Enter how much of your Capital Gains Tax Annual Exempt Amount you want to use, like 0 or 12,000
</span>
```

AndyHWChung: ah ok, then - that's ok then. We just need to have a think about whether or not should ever have an error summary that's empty. I don't think we should if this method is to specifically test for form errors

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/exemptionandlosses/ExemptionAndLossesControllerSpec.scala
```diff
@@ -1419,11 +1420,19 @@ class ExemptionAndLossesControllerSpec
...
         doc.select("#error-summary-display > ul > li > a").text() shouldBe messageFromMessageKey(
           expectedErrorMessageKey,
           errorArgs: _*
         )
+
         doc.title() should startWith("Error:")
+
+        val errorSummary     = doc.select(s"""a[href="#$key"]""").text()
+        val inlineErrorMsg  = doc.select(s"""span[id="$key-inline-error"]""").text()
+        val errorSummaryMsg = if (errorSummary.nonEmpty) "Error: " + errorSummary else errorSummary
```
AndyHWChung, line 1433:
```scala
errorSummary.nonEmpty
```
should never be true right since we should always have an error summary?

rambabu-posa: I found 4 tests were failing without that logic. Okay will check them again

PR #432, opened 2020-04-16
[CPDP-1481] - Added new attribute for input

modivilioglu:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/components/file_input.scala.html
```diff
@@ -87,6 +88,7 @@ <h1 class="govuk-label-wrapper">
   type="file"
   name="@fieldId"
   value=""
+  is-mandatory-upload="@isMandatoryUpload"
```
nubz, line 91: for attributes like this it would be better to use the `data-` prefix as data attributes will validate in HTML checks and browsers whereas unsupported attribute types may cause problems in some browsers and screenreaders. So maybe we could call the attribute `data-is-mandatory-upload` ?

PR #444, opened 2020-04-21
CPDP-1514: Unit tests fixes for Exemptionandlosses Journey

rambabu-posa:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/UserType.scala
```diff
@@ -18,6 +18,7 @@ package uk.gov.hmrc.cgtpropertydisposalsfrontend.models
 
 import julienrf.json.derived
 import play.api.libs.json.OFormat
+//import cats.Eq
```
AndyHWChung, line 21: can we remove this if it's not needed?

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/UserType.scala
```diff
@@ -28,6 +29,8 @@ object UserType {
   final case object NonGovernmentGatewayUser extends UserType
   final case object Agent extends UserType
 
+  //implicit val eq: Eq[UserType] = Eq.fromUniversalEquals
```
AndyHWChung, line 32: and this

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/UserType.scala
rambabu-posa: Will work refactoring these two specs separately (4 tests are ignored because of agents journey is missing some logic)
ViewReturnControllerSpec
 2. CheckAllAnswersAndSubmitControllerSpec

PR #449, opened 2020-04-22
Unit tests coverage

rambabu-posa:


test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/CheckAllAnswersAndSubmitControllerSpec.scala
```diff
@@ -161,7 +177,8 @@ class CheckAllAnswersAndSubmitControllerSpec
 
         "display the page" when {
 
-          def test(sessionData: SessionData, expectedTitleKey: String, userType: Option[UserType]): Unit = {
+          def test(sessionData: SessionData, expectedTitleKey: String, userType: Option[UserType], isATrust: Boolean)
```
AndyHWChung, line 180: do you need the `isATrust` flag here - shouldn't what we had before work with `ompleteFillingOutReturn.subscribedDetails.isATrust`? That is essentially what you're passing in anyway

rambabu-posa: It works for first test as it's just using existing `completeFillingOutReturn`. But does not for next test. as we are modifying `completeFillingOutReturn` with  `userType'

rambabu-posa: it took me time to figure out this issue

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/exemptionandlosses/ExemptionAndLossesControllerSpec.scala
```diff
@@ -1809,43 +1809,4 @@ object ExemptionAndLossesControllerSpec extends Matchers {
     )
   }
 
-  def validateExemptionAndLossesCheckYourAnswersPage(
```
AndyHWChung, line 1812: wasn't this being called in `ExemptionAndLossesControllerSpec` and the spec for the CheckAllAnswersController as well? If they weren't they probably should have been to check that the correct content is being displayed

rambabu-posa: Nope, just copied same function and added one extra param "isAnAgent: Boolean".
You can see that new function just above.
All are using this new function.

PR #471, opened 2020-04-30
cpdp-1593: removed upscan upload status

AndyHWChung:


mikail-khan:
Some of this code is still required - I have incorporated what is needed in my PR. Closing this PR as it will be address there.

PR #474, opened 2020-05-01
CPDP-1510: Single disposal triage content changes

rambabu-posa:


test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/triage/SingleDisposalsTriageControllerSpec.scala
```diff
@@ -2746,10 +3195,115 @@ class SingleDisposalsTriageControllerSpec
...
+              checkContent(document)
+              checkPrepopulatedContent(document)
+            }
+          )
+        }
+      }
+    }
+  }
+
+  def displayIndividualPersonalRepTriagePageBehaviorIncompleteJourney(performAction: () => Future[Result])(
```
AndyHWChung, line 3253: this function looks exactly the same as `displayIndividualCapacitorTriagePageBehaviorIncompleteJourney` just above - it makes sense to have a single method with a more generic name or if you wanted a method for personal reps and another for capacitors, it only makes sense to have separate ones if they actually do something meaningfully different.

rambabu-posa: I tried to use same but getting duplicate test error. Will try to use same method

AndyHWChung:
LGTM

PR #496, opened 2020-05-06
[CPDP-1522] - Content changes for Initial Gain or Loss

modivilioglu:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/initialgainorloss/initial_gain_or_loss.scala.html
```diff
@@ -92,28 +106,21 @@
...
-
-@title = @{messages(s"$key.title")}
-
-@mainTemplate(title = title, userType = request.userType) {
-@backLinkComponent(backLink)
-@caption(messages("initialGainOrLoss.caption"))
-@if(hasErrors) {
-    @errorSummary(form)
+@title = @{
+    messages(s"$key.title$userKey")
```
AndyHWChung, line 110: the convention we've used everywhere else is `key.userKey.x` - it'd be more consistent to go with that approach

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/initialgainorloss/InitialGainOrLossController.scala
```diff
@@ -174,7 +197,7 @@ object InitialGainOrLossController {
           }
       }
 
-    val formatter = ConditionalRadioUtils.formatter("initialGainOrLoss")(
+    val formatter = ConditionalRadioUtils.formatter(s"initialGainOrLoss", Some(s"initialGainOrLoss$userKey"))(
```
AndyHWChung, line 200: we've handled different user types at the template level for all the other pages - could we stick to that pattern for consistency?

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/initialgainorloss/InitialGainOrLossControllerSpec.scala
```diff
@@ -124,35 +130,85 @@ class InitialGainOrLossControllerSpec
...
+        )
+      }
     }
+    def generateTrustSubscribedDetails() =
+      sample[SubscribedDetails].copy(name = Left(sample[TrustName]))
+
+    def generateIndividualSubscribedDetails() =
+      sample[SubscribedDetails].copy(name = Right(sample[IndividualName]))
+
+    def getTriageAnswersWithPersonalRepresentative() =
```
AndyHWChung, line 204: is there a reason these method are `get` but the ones above are `generate`?

modivilioglu: my bad, fixed.

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/initialgainorloss/InitialGainOrLossControllerSpec.scala
```diff
@@ -107,9 +113,11 @@ class InitialGainOrLossControllerSpec
             messageFromMessageKey(
               "initialGainOrLoss.title"
             ), { doc =>
-              doc.select(".govuk-caption-xl").html()  should include("Initial gain or loss")
-              doc.select("#initialGainOrLoss").html() should include("Did you make an initial gain or loss?")
-              doc.select("#back").attr("href")        shouldBe returns.routes.TaskListController.taskList().url
+              doc.select(".govuk-caption-xl").html() should include("Initial gain or loss")
+              doc.select("#initialGainOrLoss").html() should include(
+                messageFromMessageKey(s"initialGainOrLoss.title")
```
AndyHWChung, line 118: no need for the `s` here

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/initialgainorloss/InitialGainOrLossControllerSpec.scala
```diff
@@ -345,9 +446,42 @@ class InitialGainOrLossControllerSpec
...
+        )
+
+        val testData = List(
+          ("", individualFillingOutReturn, false),
+          (".trust", trustFillingOutReturn, false),
+          (".personalRep", personalRepresentativeFillingOutReturn, false),
+          (".capacitor", capacitorFillingOutReturn, false),
+          (".agent", individualFillingOutReturn, true)
+        )
+        testData.map(keyWithJourneyStatus => test(keyWithJourneyStatus._1, keyWithJourneyStatus._2, keyWithJourneyStatus._3))
```
AndyHWChung, line 472: it makes more sense to use `foreach` instead of `map` here and elsewhere in this spec where you're doing a similar thing - using `map` you end up with a `List[Unit]`, but we would ideally end up with just `Unit` since `List[Unit]` doesn't mean anything useful

modivilioglu: Done.

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/initialgainorloss/InitialGainOrLossControllerSpec.scala
modivilioglu: Done.

PR #519, opened 2020-05-09
CPDP-1520: Capacitor acquisition content changes

rambabu-posa:


conf/messages
```diff
@@ -1176,139 +1176,256 @@ acquisitionDetails.caption=Purchase or acquisition details
...
+acquisitionPriceBought.personalRep.error.tooLarge=The market value of the property must be an amount of money between 1 and 50,000,000,000
+
 acquisitionPriceBought.error.tooManyDecimals=How much you paid for the property must be an amount of money with pence as 2 decimal places or less, like 160,000 or 23,040.50
 acquisitionPriceBought.agent.error.tooManyDecimals=How much your client paid for the property must be an amount of money with pence as 2 decimal places or less, like 160,000 or 23,040.50
 acquisitionPriceBought.trust.error.tooManyDecimals=How much the trust paid for the property must be an amount of money with pence as 2 decimal places or less, like 160,000 or 23,040.50
+acquisitionPriceBought.capacitor.error.tooManyDecimals=How much the person paid for the property must be an amount of money with pence as 2 decimal places or less, like 160,000 or 23,040.50
+acquisitionPriceBought.personalRep.error.tooManyDecimals=The market value of the property must be an amount of money with pence as 2 decimal places or less, like 160,000 or 23,040.50
+
+
+acquisitionPriceNotBought.title=What was the market value of the property on {0}?
```
AndyHWChung, line 1354: shouldn't this be `What was the market value of the property when it was acquired?` This looks like the title that should go on the rebasedAcquisitionPrice page

rambabu-posa: I think so, but not sure as its existing message. We have to check with Jean

PR #521, opened 2020-05-11
Cpdp 1609 remove ga scripts

nubz:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/govuk_wrapper.scala.html
```diff
@@ -62,31 +62,15 @@
...
+        window.jsConfig.timed_out_url = "@{appConfig.ggTimedOutUrl}";
+        window.jsConfig.logout_url = "@{if(withAccount) appConfig.accountSignOutUrl else appConfig.ggSignOut}";
+      }
     </script>
-
-    @if(appConfig.gaUserTypeDimension.nonEmpty) {
-      <script>
-        window.jsConfig.gaCustomDimensions.push({
-          name: "@{appConfig.gaUserTypeDimension}",
-          value: "@{userType match {
```
AndyHWChung, line 72: how are we going to tag events with the correct user type now? Have we lost that capability now?

nubz: this method was explicitly for sending via the GA script set up with our token - there is no token and no GA reporting script anymore so this is not going to go anywhere. Jaspreet is checking with community about custom dimension reporting (sounds like it might be data layer values - so we will most likely still need to propagate userType through to govuk_wrapper as we are doing just not report it to GA like this but do something else with it) - I expect this to be another ticket though

AndyHWChung:
LGTM

PR #538, opened 2020-05-18
CPDP-1459: Single indirect disposal step3 tasklist

rambabu-posa:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/yeartodatelliability/YearToDateLiabilityController.scala
```diff
@@ -130,6 +130,11 @@ class YearToDateLiabilityController @Inject() (
           f(s, r, IncompleteNonCalculatedYTDAnswers.empty)
         )(f(s, r, _))
 
+      case Some((s, r @ FillingOutReturn(_, _, _, d: DraftSingleIndirectDisposalReturn))) =>
```
AndyHWChung, line 133: can we update the spec of this controller to sometimes use `DraftSingleIndirectDisposalReturn` where it's currently using `DraftMultipleDisposalsReturn`? That's the approach I took with some of the other controllers - rather than writing duplicate tests for the different draft return type I swapped them out to get some coverage of both types

AndyHWChung:
LGTM

PR #540, opened 2020-05-18
cpdp-1466: fixed  back links

AndyHWChung:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/AddressController.scala
```diff
@@ -77,7 +77,8 @@ trait AddressController[A <: AddressJourneyType] {
   protected val selectAddressSubmitCall: Call
   protected val continueCall: Call
 
-  protected val enterPostcodePageBackLink: A => Call = _ => isUkCall
+  protected lazy val enterUkAddressBackLinkCall: A => Call = backLinkCall
+  protected val enterPostcodePageBackLink: A => Call       = _ => isUkCall
```
mikail-khan, line 81: why is one lazy and the other isn't?

AndyHWChung: I made the new one lazy cos I wanted it lazy in the controller I'm overwriting it in - I can take a look at the non lazy one to see what's going on there

AndyHWChung: though maybe the one I added in doesn't need to be lazy after all...

AndyHWChung: turns out it didn't need to be lazy - tested ok with it being non-lazy

PR #546, opened 2020-05-19
CPDP-1555: Multiple indirect disposal tasklist

rambabu-posa:


conf/messages
```diff
@@ -969,10 +969,10 @@ task-list.check-and-send-return.link=Check and send return, pay any tax due
 task-list.upload-supporting-docs.link=Upload Your Supporting Documents
 
 
-task-list.enter-example-property-address.link=Enter details of one property
+task-list.enter-example-property-address.link=Enter details of one company
```
AndyHWChung, line 972: the old values still need to be there when it's a multiple disposal journey which isn't indirect disposals

rambabu-posa: oh okay sure, will fix them soon

conf/messages
```diff
@@ -968,11 +968,12 @@ task-list.ytd-liability.link=Enter Capital Gains Tax liability so far this tax y
 task-list.check-and-send-return.link=Check and send return, pay any tax due
 task-list.upload-supporting-docs.link=Upload Your Supporting Documents
 
-
 task-list.enter-example-property-address.link=Enter details of one property
+task-list.indirect.enter-example-property-address.link=Enter details of one company
 
 returns.property-address.singleDisposal.caption=Property address
 returns.property-details.multipleDisposals.caption=Details of one property disposal
+returns.indirect.property-details.multipleDisposals.caption=Details of one company
```
AndyHWChung, line 976: for this ticket you shouldn't need to worry about the caption - that'll be tackled in CPDP-1557. The changes you've done to make it happen may not be the way it's done when that ticket is done since the pages may sit in another controller with new pages.

rambabu-posa: Oh as story description is saying to work on that one too, I fixed it for all views.
so do I need to work on only tasklist message?
I did lot of stuff for changing caption..

AndyHWChung: I think that was a late addition that may not have been needed since it's covered by a subsequent ticket. The way we do it when we pick up that ticke may not be the way you've done it just now, so it would be safer to do the changes in that ticket.

AndyHWChung:
LGTM

PR #547, opened 2020-05-20
CPDP-1467: handle upscan errors

mikail-khan:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/supportingevidence/SupportingEvidenceController.scala
```diff
@@ -246,48 +325,70 @@ class SupportingEvidenceController @Inject() (
...
-  def uploadSupportingEvidenceError(): Action[AnyContent] = authenticatedActionWithSessionData { implicit request =>
-    errorHandler.errorResult()
-  }
+  def handleUpscanErrorRedirect(): Action[AnyContent] =
+    authenticatedActionWithSessionData {
+      Redirect(routes.SupportingEvidenceController.documentDidNotUpload())
+    }
 
-  def uploadSupportingEvidenceVirusCheck(uploadReference: UploadReference): Action[AnyContent] =
+  def documentDidNotUpload(): Action[AnyContent] =
```
AndyHWChung, line 341: could we get a unit test for this endpoint and the one below it? If you use `checkPageIsDisplayed` you can at least check the page title and the href of the green buttons. We'll eventually get automatic checks that there aren't any missing message keys as well with that method.

mikail-khan: added

AndyHWChung:
LGTM

PR #549, opened 2020-05-21
[CPDP-1631] - Add tests to discover missing messages

modivilioglu:


test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/ControllerSpec.scala
```diff
@@ -24,17 +24,54 @@ import org.jsoup.Jsoup
...
+@Singleton
+class TestDefaultMessagesApiProvider @Inject() (
+  environment: Environment,
+  config: Configuration,
+  langs: Langs,
+  httpConfiguration: HttpConfiguration
+) extends DefaultMessagesApiProvider(environment, config, langs, httpConfiguration) {
+
+  override lazy val get: MessagesApi = {
+    new TestMessagesApi(
```
AndyHWChung, line 65: you don't need `@Inject` in `TestMessagesApi` if you're using `new` to create an instance of it. Also, you won't need default arguments if you're always giving concrete values here when you use it

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/ControllerSpec.scala
```diff
@@ -107,7 +145,8 @@ trait ControllerSpec extends WordSpec with Matchers with BeforeAndAfterAll with
     status(result)           shouldBe expectedStatus
 
     val doc = Jsoup.parse(contentAsString(result))
-    doc.select("h1").text should include(expectedTitle)
+    doc.select("h1").text   should include(expectedTitle)
+    doc.select("body").text should not include ("not_found_message(")
```
AndyHWChung, line 149: it would be more helpful if the test logs said which message keys were missing. You could do something like this:
```scala
...
val regex = ???
val missingMessageKeys: List[String] = ??? // find all instances where regex is satisfied
if(missingMessageKeys.nonEmpty) fail("Message keys are missing: $missingMessageKeys") 
...
```

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/reliefdetails/ReliefDetailsControllerSpec.scala
```diff
@@ -691,9 +691,11 @@ class ReliefDetailsControllerSpec
       behave like noPrivateResidentsReliefBehaviour(() => performAction(Seq.empty))
 
       "show a form error" when {
+        type Message    = String
+        type MessageKey = String
 
         def test(data: (String, String)*)(
-          expectedErrorMessageKey: String
+          expectedErrorMessageOrKey: Either[MessageKey, Message]
```
AndyHWChung, line 698: you could just take the key as one parameter and the arguments as another parameter and pass them both to `messageFromMessageKey` - that way you can avoid an `Either`

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/ControllerSpec.scala
modivilioglu: Oh yes, I tried to inject the TestMessagesApi first, then when I used the Provider, I forgot to delete it. My bad

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/reliefdetails/ReliefDetailsControllerSpec.scala
modivilioglu: Done.

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/ControllerSpec.scala
modivilioglu: Done.

PR #550, opened 2020-05-22
CPDP-1468: Single Indirect Disposal details content changes

rambabu-posa:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/disposaldetails/DisposalDetailsController.scala
```diff
@@ -504,15 +517,18 @@ class DisposalDetailsController @Inject() (
           withDisposalMethod(state) {
             case (disposalMethod) =>
               answers match {
-                case IncompleteDisposalDetailsAnswers(None, _, _) =>
+                case IncompleteDisposalDetailsAnswers(None, _, _) if isIndirectDisposal(state) =>
```
AndyHWChung, line 520: will this ever happen if you're always populating the share of the property for indirect disposals?

rambabu-posa: Yes I think so. Thats why Indirect disposal is redirecting to 'disposal price' page directly.

AndyHWChung: I think it's redirecting to that page because when we get the the cya endpoint we always start off with `IncompleteAnswers(Some(_),...)` so it's always skipping this case. It looks like this case is redundant - if you take it out and try it on the browser, does it not work still?

rambabu-posa: Yes you are right.

AndyHWChung:
LGTM

PR #552, opened 2020-05-22
[CPDP-1468] - Added price page test cases for indirect disposals

modivilioglu:


rambabu-posa:
As Im using different approach for these unit tests, am going to close this PR without merging.

PR #553, opened 2020-05-27
[CPDP-1765] - Change file expiry time from 7 days to 5 days

modivilioglu:


AndyHWChung:
as well as `application.conf` could you please also check app-config-base, app-config-development, etc.?

AndyHWChung:
LGTM

PR #555, opened 2020-05-27
CPDP-1625: ur banner with config and js

nubz:


conf/application.conf
```diff
@@ -249,4 +249,9 @@ mixed-use {
   enabled = true
 }
 
+user-recruitment-banner {
+  enabled = true
```
AndyHWChung, line 253: do we need to set this to false in prod for now or are we ok with enabling it in prod straight away?

nubz: as i understand it there is a push to get this live asap

nubz: just confirmed with Ashleigh - the sooner the better and the link is good to go now

AndyHWChung:
LGTM

PR #569, opened 2020-06-01
CPDP-1624: draft and sent labels for indirect

nubz:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/account/components/draft_list_item.scala.html
```diff
@@ -65,28 +64,54 @@
...
 }
+
 @dueDate = @{completionDate._1.map(d => d.plusDays(30L))}
 
 @getAddressLines(a: Address) = @{
     a match {
-        case u: UkAddress => s"${u.line1}, ${u.postcode.value}"
-        case n: NonUkAddress => s"${n.line1}, ${n.country.name.getOrElse("")}"
+      case u: UkAddress => s"${u.line2 match {
+        case Some(l2) => s"${l2}, "
```
AndyHWChung, line 83: `s"${l2}"` can be simplified to just `l2`. The same goes for the things below as well.

AndyHWChung: There's also a neater way to do this - you could do something like this:
https://github.com/hmrc/cgt-property-disposals-frontend/blob/master/test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/address/CompanyDetailsControllerSpec.scala#L1235
then do `.mkString(", ")` on the resulting `List[String]`. That method is probably generic enough to go into an implicit class on `Address` in a method called `toAddressLineList()` or something - then you could use it like this:
```scala
@getAddressLines(a: Address) = @{ 
  a.toAddressLineList().tailOption.getOrElse(List.empty).mkString(", ") 
}
```

nubz: > `s"${l2}"` can be simplified to just `l2`. The same goes for the things below as well.

did need the comma when value is there it was actually s"${l2}, "  but will have a go at the generic method so we can use in both templates and also the spec

AndyHWChung:
LGTM

PR #570, opened 2020-06-02
[CPDP-Warning-Cleanup] - Cleaning up warnings

modivilioglu:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/representee/RepresenteeController.scala
```diff
@@ -222,7 +222,7 @@ class RepresenteeController @Inject() (
 
   def confirmPerson(): Action[AnyContent] =
     authenticatedActionWithSessionData.async { implicit request =>
-      withCapacitorOrPersonalRepresentativeAnswers(request) { (representativeType, journey, answers) =>
+      withCapacitorOrPersonalRepresentativeAnswers(request) { (_, journey, answers) =>
```
mikail-khan, line 225: Looks like the `WithCapacitorOrPersonalRepresentativeAnswers` needs to change to remove a function parameter. If so take it out there rather than applying the underscore

mikail-khan:
good job Mehmet :smile: - let's add in the other compiler flags and give this another whirl.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/representee/RepresenteeController.scala
modivilioglu: No, that parameter is used on some calls in the file.

PR #571, opened 2020-06-02
CPDP-1610: non-resident attempting to change to UK address

mikail-khan:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/address/PropertyDetailsController.scala
```diff
@@ -945,7 +946,7 @@ class PropertyDetailsController @Inject() (
   protected lazy val enterNonUkAddressCall: Call                          = enterPostcodeCall
   protected lazy val enterNonUkAddressSubmitCall: Call                    = enterPostcodeCall
   protected lazy val backLinkCall: FillingOutReturnAddressJourney => Call = _ => enterPostcodeCall
-
+  protected lazy val exitPageCall: Call                                   = routes.PropertyDetailsController.showExitPage()
```
mikail-khan, line 949: wasn't sure about this - we don't really need the exit page for this part of the service functionality

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/AddressControllerSpec.scala
```diff
@@ -238,14 +239,79 @@ trait AddressControllerSpec[A <: AddressJourneyType]
       }
     }
 
+    "redirect to the exit page" when {
+      "a registered without id user attempts to change their address to a UK address" in {
+        validJourneyStatus match {
+          case j: AddressJourneyType.ManagingSubscription.SubscribedAddressJourney =>
+            val subscribedDetails: SubscribedDetails = sample[SubscribedDetails].copy(registeredWithId = false)
```
mikail-khan, line 246: This is all smelly - we have to leak the exact JournetStatus type into the AddressControllerSpec but I can't see any other way to do it

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/AddressController.scala
```diff
@@ -429,4 +445,19 @@ trait AddressController[A <: AddressJourneyType] {
...
+          case Onboarding.RegistrationReadyAddressJourney(_)              => Some(false)
+          case Onboarding.IndividualSupplyingInformationAddressJourney(_) => Some(false)
+          case Onboarding.SubscriptionReadyAddressJourney(_)              => Some(true)
+        }
+      case subscription: AddressJourneyType.ManagingSubscription =>
+        subscription match {
+          case ManagingSubscription.SubscribedAddressJourney(subscribed) =>
+            Some(subscribed.subscribedDetails.registeredWithId)
+        }
+      case _: AddressJourneyType.Returns                         => Some(true)
```
AndyHWChung, line 461: this might not be `true` since you can do a return when you've registered without id. You can avoid having to consider whether or not someone regesiterd with or without id where you don't need to though if you take a more localised approach to displaying the exit page. So if you had:
```scala
def ukAddressNotAllowed(): Action[AnyContent] = Action {
  exit_page()
} 
```
in some controller somewhere then inside `RegistrationChangeAddressController.scala` and `RegistrationEnterAddressController.scala` you could just have:
```
override def enterPostcode(): Action[AnyContent] = Action { _ => Redirect(ukAddressNotAllowed)}
override def enterPostcodeSubmit(): Action[AnyContent] = Action { _ => Redirect(ukAddressNotAllowed)}
override def ukAddress(): Action[AnyContent] = Action { _ => Redirect(ukAddressNotAllowed)}
override def ukAddressSubmit(): Action[AnyContent] = Action { _ => Redirect(ukAddressNotAllowed)}
```
and in the `SubscribedChangeAddressController.scala` the same thing expect you guard the redirect with a check on whether they registered with or without id.  Then you don't need this method or any changes to the `AddressController` or the `AddresControllerSpec` - you'd only need changes to the specific controllers and their specs. 
Also, having two separate endpoints defined for the same url in the routes file won't work since you'll always be redirect to one controller and the states won't match up in some of the journeys.

AndyHWChung: There's no natural controller for the "global" exit page, so maybe it could just go in the `StartController` with the above solution. I think there's a couple of exit pages there already.

mikail-khan: I did look at putting it in the `StartController` first but this exit page is not a global one but rather a specific one concerning addresses which is why I placed it under the address the package. That's my interpretation of it. Is that right?

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/AddressController.scala
```diff
@@ -429,4 +445,19 @@ trait AddressController[A <: AddressJourneyType] {
       case _                                 => None
     }
 
+  private def registeredWithId(journey: AddressJourneyType): Option[Boolean] =
```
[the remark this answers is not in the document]
mikail-khan: For 
`scala
final case class SubscribedDetails(
  name: Either[TrustName, IndividualName],
  emailAddress: Email,
  address: Address,
  contactName: ContactName,
  cgtReference: CgtReference,
  telephoneNumber: Option[TelephoneNumber],
  registeredWithId: Boolean
)
```
will always have it populated.
But for the other scenarios, it won't be there which is the reason why this is an `Option`

File: app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/AddressController.scala (hunk 3 above)
Comment (mikail-khan) [reply app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/AddressController.scala:461]: > this might not be `true` since you can do a return when you've registered without id. You can avoid having to consider whether or not someone regesiterd with or without id where you don't need to though if you take a more localised approach to displaying the exit page. So if you had:
> 
> ```scala
> def ukAddressNotAllowed(): Action[AnyContent] = Action {
>   exit_page()
> } 
> ```
> 
> in some controller somewhere then inside `RegistrationChangeAddressController.scala` and `RegistrationEnterAddressController.scala` you could just have:
> 
> ```
> override def enterPostcode(): Action[AnyContent] = Action { _ => Redirect(ukAddressNotAllowed)}
> override def enterPostcodeSubmit(): Action[AnyContent] = Action { _ => Redirect(ukAddressNotAllowed)}
> override def ukAddress(): Action[AnyContent] = Action { _ => Redirect(ukAddressNotAllowed)}
> override def ukAddressSubmit(): Action[AnyContent] = Action { _ => Redirect(ukAddressNotAllowed)}
> ```
> 
> and in the `SubscribedChangeAddressController.scala` the same thing expect you guard the redirect with a check on whether they registered with or without id. Then you don't need this method or any changes to the `AddressController` or the `AddresControllerSpec` - you'd only need changes to the specific controllers and their specs.
> Also, having two separate endpoints defined for the same url in the routes file won't work since you'll always be redirect to one controller and the states won't match up in some of the journeys.

I'll give this some thought. Not sure if I understand it yet.

Comment (mikail-khan) [reply app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/AddressController.scala:461]: > this might not be `true` since you can do a return when you've registered without id. You can avoid having to consider whether or not someone regesiterd with or without id where you don't need to though if you take a more localised approach to displaying the exit page. So if you had:
> 
> ```scala
> def ukAddressNotAllowed(): Action[AnyContent] = Action {
>   exit_page()
> } 
> ```
> 
> in some controller somewhere then inside `RegistrationChangeAddressController.scala` and `RegistrationEnterAddressController.scala` you could just have:
> 
> ```
> override def enterPostcode(): Action[AnyContent] = Action { _ => Redirect(ukAddressNotAllowed)}
> override def enterPostcodeSubmit(): Action[AnyContent] = Action { _ => Redirect(ukAddressNotAllowed)}
> override def ukAddress(): Action[AnyContent] = Action { _ => Redirect(ukAddressNotAllowed)}
> override def ukAddressSubmit(): Action[AnyContent] = Action { _ => Redirect(ukAddressNotAllowed)}
> ```
> 
> and in the `SubscribedChangeAddressController.scala` the same thing expect you guard the redirect with a check on whether they registered with or without id. Then you don't need this method or any changes to the `AddressController` or the `AddresControllerSpec` - you'd only need changes to the specific controllers and their specs.
> Also, having two separate endpoints defined for the same url in the routes file won't work since you'll always be redirect to one controller and the states won't match up in some of the journeys.

One of the problems with your proposed solution is the back-link. The way I have it at the moment is that the back-link can be easily set inside the respective subtype of the AddressController type. 

What you are proposing is a general handler for this. With that we cannot easily go back to the page the user can from unless we thread through from which page the redirect to the exit-page was triggered.

Comment (mikail-khan) [reply app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/AddressController.scala:461]: I could pass a URL (string) to the controller method:

```scala
def ukAddressNotAllowed(backLink :String)
```

and then feed that into the general template.

I'll go with that and raise another PR.

AndyHWChung: oh yeah, that's a good point - it's a trade off then between having to cater for all journey types in the `AddressController` trait and being able to determine the back link in some generic exit page . The latter would overall be more annoying to do I think - but for the first one, we need different URL's for the exit page depending on the journey type then. One way could be to  tag on `exit/cannot-use-service-non-resident-uk-address` to the existing urls that exist for the sections like:
```
/registration/create-account/exit/cannot-use-service-non-resident-uk-address
/registration/create-account/exit/cannot-use-service-diferent-non-resident-uk-address
/account-home/manage-account/exit/cannot-use-service-non-resident-uk-address
```
We'd have to get that double checked though

mikail-khan: that's how it works at the moment - each subtype of the `AddressController` trait has it's own exit url. If you have another look at the router files, I made those changes in there.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/address/PropertyDetailsController.scala
AndyHWChung: maybe a way to avoid having to do this where we don't need it is to have the type of this `Option[Call]` and default it to `None`, then override it where we do need it. Then we could avoid having to figure out whether or not it's register with or without id in the `AddressController` since you'd just do something like this if the person chose yes to the is uk question:
```scala
 Redirect(ukAddressNotAllowedExitPageCall.getOrElse(enterPostcodeCall)
```
I put a more descriptive name in there as well.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/AddressController.scala
mikail-khan: scratch that, I see what you mean. Yeah we would need to change the urls as you described above. Who can do that? Jean I suppose.

mikail-khan: If we do make the URL changes, then I think this is good to go. Basically your solution approaches the problem from the opposite angle but with the price that we have to feed the backLink url through. 

We need to make a decision :smile:

AndyHWChung: there's a security risk with having a URL as a query parameter as another URL - I can't remember exactly when but I remember it being raised at some point on another project. What might be easiest then is to go with what you've got, but maybe think about the `Option[Call]` approach in my other comment to reduce the number of cases we have to worry about with this addition

AndyHWChung: the risk was something about people being able to share a url in an email or something which looked like a legitimate tax platform url but make the back link point to something malicious.

mikail-khan: That's fine with me. I'll review the review comments and make the changes.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/address/PropertyDetailsController.scala
mikail-khan: changed to make the call optional

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/AddressController.scala
mikail-khan: fixed as agreed

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/AddressController.scala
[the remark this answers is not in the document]
AndyHWChung: are you ever returning `None` in the function or am I missing something? It looks like all the cases point to `Some`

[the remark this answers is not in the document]
mikail-khan: Yes, I can change the return type to `Boolean`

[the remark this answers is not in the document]
mikail-khan: fixed

AndyHWChung:
LGTM

PR #574, opened 2020-06-02
Cpdp 1550 fixes

sam-miele:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/acquisitiondetails/AcquisitionDetailsController.scala
```diff
@@ -543,10 +543,12 @@ class AcquisitionDetailsController @Inject() (
...
-                    rebasingEligibilityUtil.getDisplayRebasingCutOffDate(
-                      assetType,
-                      wasUkResident
-                    ),
+                    rebasingEligibilityUtil
+                      .getDisplayRebasingCutOffDate(
+                        assetType,
+                        wasUkResident
+                      )
+                      .minusDays(1),
```
AndyHWChung, line 551: would it make more sense not take a day away from here since this param is supposed to be the actual cut off date - you could just minus a day in the message where you need the day taken off. Is that the only place you need a day taken off?

sam-miele: Sure. I will make that change. I'll have a look for usages passed in here and update them if needed

AndyHWChung:
LGTM

PR #576, opened 2020-06-03
Cpdp 1835

modivilioglu:
- For the FE I've cleaned up the warnings generated by the flags (there was 123 in the beginning), except the one I explain below
- ```/Users/user/Documents/workspaces/cgt-property-disposals-frontend/app/uk/gov/hmrc/cgtpropertydisposalsfrontend/config/ErrorHandler.scala:47:5: parameter value userType in method errorResult is never used```
On the above warning, if I remove there will be conflict with another method with the same name, and once I change the signature - since not private - many places will fail in the code.
This might be left with a suppresswarnings equivalent.

### DIFF
--- a/app/uk/gov/hmrc/cgtpropertydisposalsfrontend/config/ErrorHandler.scala
+++ b/app/uk/gov/hmrc/cgtpropertydisposalsfrontend/config/ErrorHandler.scala
@@ -41,14 +41,13 @@ class ErrorHandler @Inject() (
   )(implicit
     request: Request[_]
   ): Html =
-    error_template(None, pageTitle, heading, message)
+    error_template(pageTitle, heading, message)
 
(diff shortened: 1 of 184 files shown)


mikail-khan:
> * For the FE I've cleaned up the warnings generated by the flags (there was 123 in the beginning), except the one I explain below
> * `/Users/user/Documents/workspaces/cgt-property-disposals-frontend/app/uk/gov/hmrc/cgtpropertydisposalsfrontend/config/ErrorHandler.scala:47:5: parameter value userType in method errorResult is never used`
>   On the above warning, if I remove there will be conflict with another method with the same name, and once I change the signature - since not private - many places will fail in the code.
>   This might be left with a suppresswarnings equivalent.

that sounds a like massive code smell @modivilioglu  - I will look into. Thanks for exposing it.

PR #588, opened 2020-06-10
CPDP-2021: Change Property Address label for indirect disposals

rambabu-posa:


AndyHWChung:
were there no breaking unit tests? It'd be nice to get a unit test which tested this behaviour when it's indirect disposals

rambabu-posa:
No, all unit tests are working as is

rambabu-posa:
Sure will try to write unit tests for those two controllers

PR #594, opened 2020-06-12
CPDP-1840: Fix for DES postcode lookup issue

rambabu-posa:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/services/UKAddressLookupService.scala
```diff
@@ -87,7 +86,22 @@ class UKAddressLookupServiceImpl @Inject() (
     postcode: Postcode,
     filter: Option[String]
   ): Either[String, AddressLookupResult] = {
-    def toAddress(a: RawAddress): Either[String, Address] = {
+
+    def validateAddressFields(address: UkAddress): Boolean = {
+      val addressLineRegex = "^[A-Za-z0-9 \\-,.&'/]{0,35}$".r.pattern.asPredicate()
```
AndyHWChung, line 91: you could put these regex outside of the function to avoid creating a load a temporary vlaues which are the same thing everything this function is being called. Since it's the same regex that is used in `Address` and `Postcode`, you could possible make those available in those objects to reuse here

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/services/UKAddressLookupServiceImplSpec.scala
```diff
@@ -119,6 +119,120 @@ class UKAddressLookupServiceImplSpec extends WordSpec with Matchers with MockFac
           ).isLeft shouldBe true
         }
 
+        "address line1 is more than 35 characters" in {
```
AndyHWChung, line 122: the service isn't returning an error in this case - we should have a different section of the spec which is something like:
```
  "filter out invalid addresses" when { 
    "address line1 is more than 35 characters" in { ...}
  }
```

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/services/UKAddressLookupServiceImplSpec.scala
```diff
@@ -119,6 +119,120 @@ class UKAddressLookupServiceImplSpec extends WordSpec with Matchers with MockFac
...
+              |  }
+              |]
+              |""".stripMargin
+          )
+
+          mockLookupAddress(postcode)(Right(HttpResponse(200, Some(json))))
+
+          await(
+            service.lookupAddress(postcode, None).value
+          ).exists(_.addresses.isEmpty) shouldBe true
```
AndyHWChung, line 142: testing like this would result in error messages that are more helpful if something goes wrong with the test:
```
await(service.lookupAddress(postcode, None).value) shouldBe Right(AddressLokupResult(...,..., List.empty))
```
Otherwise, we'd see `false was not equal to true` which as helpful as seeing the whole result itself

AndyHWChung:
LGTM

PR #602, opened 2020-06-18
CPDP-1803: Fix for before 1-1-1900 dates and tests

rambabu-posa:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/TimeUtils.scala
```diff
@@ -88,7 +88,8 @@ object TimeUtils {
         key: String,
         data: Map[String, String]
       ): Either[Seq[FormError], LocalDate] = {
-        val result = for {
+        val minimumDate = LocalDate.of(1900, 1, 1)
```
AndyHWChung, line 91: we could move this out of the `def` to avoid creating a new `LocalDate` which is the same value every time this function is called

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/TimeUtils.scala
```diff
@@ -99,6 +100,8 @@ object TimeUtils {
                    .flatMap(date =>
                      if (maximumDateInclusive.exists(_.isBefore(date)))
                        Left(FormError(dateKey, "error.tooFarInFuture"))
+                     else if (date.isBefore(minimumDate))
```
AndyHWChung, line 103: this case should go below the `if (minimumDateInclusive.exists(_.isAfter(date)))` case

AndyHWChung:
LGTM

PR #606, opened 2020-06-22
CPDP-1800: Content changes for Property Details (Single Disposal)

rambabu-posa:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/address/enter_postcode.scala.html
```diff
@@ -29,33 +29,34 @@
...
+    representativeType: Option[RepresentativeType]
 )(implicit request: RequestWithSessionData[_], messages: Messages, appConfig: ViewConfig)
 
 @key = @{"postcode"}
 @isAgent = @{ request.userType.contains(UserType.Agent) }
 @userKey = @{
   representativeType match {
       case Some(PersonalRepresentative) => ".personalRep"
-      case Some(PersonalRepresentativeInPeriodOfAdmin) => ".personalRep"
+      case Some(PersonalRepresentativeInPeriodOfAdmin) => ".personalRepInPeriodOfAdmin"
```
AndyHWChung, line 59: shouldn't this be something like this?
```scala
@userKey = @{
  representativeType match {
      case Some(PersonalRepresentative) => ".personalRep"
      case Some(PersonalRepresentativeInPeriodOfAdmin)  => if(isAgent) ".agent" else ".personalRep"
      case Some(Capacitor) => ".capacitor"
      case None =>
          if (isAgent) ".agent"
          else if (isATrust) ".trust"
          else ""
  }
}
```
the content in the ticket looks like it's specifically for _agents_ of period of admin people.

rambabu-posa: we dont need separate entry for ".personalRepInPeriodOfAdmin" in messages file?

AndyHWChung: cos there's no specific message for non-agent period of admin users I assume that it would follow the same content as personal representatives. We could either reuse the same key for periodOfAdmin or create a new one and put the same content for personalRep in it. We may end up with a lot of duplicate messages though with the latter option.

AndyHWChung:
LGTM

PR #613, opened 2020-06-25
CPDP 2207

sam-miele:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/view_return.scala.html
```diff
@@ -66,9 +71,24 @@
 @amountOwed = @{returnSummary.mainReturnChargeAmount}
 
 @key= @{ "viewReturn" }
+@isAgent = @{ request.userType.contains(UserType.Agent) }
+@isATrust = @{ subscribedDetails.isATrust }
+
+@userKey = @{
+  representativeType match {
```
AndyHWChung, line 78: do you need the representative type here? Capacitors and personal reps will never see returns they've submitted on behalf of someone else - only the person being represented will see the return if they can and in that instance they'll see this page as an individual. The ticket only mentions content for individuals, agents and trusts.

PR #614, opened 2020-06-25
Cpdp 2219 reorganise messages for welsh

nubz:


conf/messages
```diff
@@ -215,10 +162,156 @@ agentsLandingPage.what-you-will-need-to-complete-the-return.p2=You may be asked
...
+timed-out.title=For your security, we signed you out
+timed-out.p1=This was because you were inactive for too long.
+timed-out.p2=To access your Capital Gains on UK property account, you’ll need to sign in.
+timed-out.button=Sign in
+
+recruitment-banner.message=Help improve GOV.UK
+recruitment-banner.dismiss=No thanks
+recruitment-banner.link=Get involved in making Government services better (opens in a new tab)
+
+subsequentReturnExit.title=You’ll need to contact HMRC
```
AndyHWChung, line 260: we're not likely to need this particular page translated - we'll have subsequent returns (hopefully) before we go live. We could send it to them anyway though just in case

conf/messages
```diff
@@ -602,7 +633,6 @@ email-whitelisting.problem.bullet4=the trial has ended
 email-whitelisting.problem.p3=You’ll need to email HMRC at <strong class="bold">ddcw-info-cgtpd-g@digital.hmrc.gov.uk</strong> to request access to the service.
 email-whitelisting.problem.p4=You should include your name.
 
-
 agent.enter-client-cgt-ref.title=What is your client’s Capital Gains Tax on UK property account number?
```
AndyHWChung, line 636: not sure if I'd class this as onboarding - this might be worth taking out into a separate `Agent` section. That's how the code is structured anyway

conf/messages
```diff
@@ -602,7 +633,6 @@ email-whitelisting.problem.bullet4=the trial has ended
 email-whitelisting.problem.p3=You’ll need to email HMRC at <strong class="bold">ddcw-info-cgtpd-g@digital.hmrc.gov.uk</strong> to request access to the service.
 email-whitelisting.problem.p4=You should include your name.
```
AndyHWChung, line 634: We don't need these messages anymore now I don't think - we don't do anything with whitelisting anymore. Not sure if there are templates lurking around somewhere which are still using these messages

conf/messages
nubz: We can't remove it from messages file until we stop using it so I put it into the new TRANSLATION NOT REQUIRED section at the top along with landing page stuff

conf/messages
nubz: moved it into new AGENT section

conf/messages
nubz: there is no usage so I deleted them

AndyHWChung:
LGTM

PR #616, opened 2020-06-25
CPDP-1800: [POA]Single Disposal Property Details message typos fixes

rambabu-posa:


conf/messages
```diff
@@ -1331,19 +1331,19 @@ property-details.multiple-disposals.guidance.capacitor.p3=Contract exchange usua
 property-details.multiple-disposals.guidance.p4=You’ll also need to upload a document later providing details of the other properties you’re including in this return.
 
 enterPostcode.returns.singleDisposal.title=Find the address of the UK property you disposed of
-enterPostcode.returns.agent.singleDisposal.title=Find the address of the UK property your client disposed of
+enterPostcode.returns.agent.singleDisposal.title=Find the address of the UK property the person disposed of
```
AndyHWChung, line 1334: does this content change apply to all agents or just agents of period of admin people?

rambabu-posa: As per story description " The user enters this user journey when they answer 'Who are you completing this return for?' with 'The estate of a person who died before disposing of a property."

I think it applies to "Agent of an Estate" only.

AndyHWChung: you need a separate user key to cover this then. I used `.personalRepInPeriodOfAdmin.agent` for mine when I had something similar - the pattern match for the user key will have to then have a case to return that value.

AndyHWChung: if it helps this is what I ended up doing: https://github.com/hmrc/cgt-property-disposals-frontend/blob/master/app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/disposaldetails/disposal_fees.scala.html#L63

conf/messages
```diff
@@ -1332,18 +1332,21 @@ property-details.multiple-disposals.guidance.p4=You’ll also need to upload a d
 
 enterPostcode.returns.singleDisposal.title=Find the address of the UK property you disposed of
 enterPostcode.returns.agent.singleDisposal.title=Find the address of the UK property your client disposed of
+enterPostcode.returns.personalRepInPeriodOfAdmin.agent.singleDisposal.title=Find the address of the UK property the person disposed of
```
AndyHWChung, line 1335: does the template return this user key already?

rambabu-posa: Nope, Im still working on that one

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/address/enter_postcode.scala.html
```diff
@@ -65,18 +65,26 @@
   }
 }
 
+@userMsgKey = @{
```
AndyHWChung, line 68: something like `titleUserMessageKey` would be more descriptive here - it's hard to tell what the difference is between `userMsgKey` and `userKey`

AndyHWChung:
LGTM

PR #651, opened 2020-07-08
CPDP-2322: cap/pr is first return page

AndyHWChung:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/representee/ChangeRepresenteeContactAddressController.scala
```diff
@@ -144,7 +144,8 @@ class ChangeRepresenteeContactAddressController @Inject() (
           complete.dateOfDeath,
           Some(newContactDetails),
           hasConfirmedPerson = true,
-          hasConfirmedContactDetails = false
+          hasConfirmedContactDetails = false,
```
mikail-khan, line 147: why is this needed as part of checking if it is a first return or not?

AndyHWChung: do you mean `hasConfirmedContactDetails`? That flag isn't required for the first return stuff, that was there before. it tells us whether or not the person has clicked confirm on the cya page where we display contact details that the capacitor or personal rep wants to use in the return

mikail-khan: ignore that - I haven't woken up yet

PR #655, opened 2020-07-13
CPDP-1618: [POA] Reliefs Message fixes

rambabu-posa:


conf/messages
```diff
@@ -2827,8 +2828,8 @@ otherReliefs.agent.title=Does your client want to claim any other Capital Gains
 otherReliefs.trust.title=Does the trust want to claim any other Capital Gains Tax reliefs?
 otherReliefs.capacitor.title=Does the person want to claim any other Capital Gains Tax reliefs?
 otherReliefs.personalRep.title=Was the person entitled to claim any other Capital Gains Tax reliefs?
-otherReliefs.personalRepInPeriodOfAdmin.agent.title=Are any other Capital Gains Tax reliefs being claimed?
-otherReliefs.personalRepInPeriodOfAdmin.title=Are any other Capital Gains Tax reliefs being claimed?
+otherReliefs.personalRepInPeriodOfAdmin.agent.title= Are any other reliefs being claimed?
```
AndyHWChung, line 2831: there's some extra white space at the start of this message and the one below - it's safer to remove them in case they end up getting rendered

nubz, line 2831: I think we have a conflict of ticket precedence here - the more recent ticket I did changed this to include Capital gains Tax in the message https://jira.tools.tax.service.gov.uk/browse/CPDP-2303

rambabu-posa: okay, we will go with 1618 first, then please apply those changes. Is that okay?

nubz: I think QA should make the call - I thought Umar was now looking at both tickets together because they affect the same messages.

rambabu-posa: Okay thanks, will revert those changes

PR #659, opened 2020-07-14
CPDP-1618: Fix for letting relief page

rambabu-posa:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/reliefdetails/ReliefDetailsController.scala
```diff
@@ -510,13 +510,13 @@ class ReliefDetailsController @Inject() (
               ) =>
             Redirect(routes.ReliefDetailsController.otherReliefs())
 
-          case IncompleteReliefDetailsAnswers(_, None, None)           =>
+          case IncompleteReliefDetailsAnswers(Some(prr), None, _) if prr > AmountInPence.zero =>
```
AndyHWChung, line 513: it'd be good to get a unit test to test this new behaviour - I assume there isn't one already since it's new

rambabu-posa: Do you want to write it in cya submit section?

AndyHWChung: yeah, there should be a bunch of tests there already which test each of the cases in the pattern match

AndyHWChung:
LGTM

PR #661, opened 2020-07-14
setting play.118n http only cookie settings to false

nubz:


conf/application.conf
```diff
@@ -266,9 +266,9 @@ enable-language-switching = true
 play.i18n {
     langs = ["en", "cy"]
     # Whether the HTTP only attribute of the cookie should be set to true
-    langCookieHttpOnly = true
+    langCookieHttpOnly = false
```
mikail-khan, line 269: remove this and for the environment config files as well

conf/application.conf
```diff
@@ -266,9 +266,9 @@ enable-language-switching = true
...
     # The lang cookie secure set to true
-    langCookieSecure = true
+    langCookieSecure = false
```
mikail-khan, line 271: remove this and for the environment config files as well

mikail-khan:
if you raise the PRs for the environment configs and get those merged first then I will merge this one so it picks up the new environment configuration

PR #668, opened 2020-07-15
CPDP-1792: use stubs to process files

mikail-khan:


conf/application.conf
```diff
@@ -105,7 +106,8 @@ microservice {
     upscan-initiate {
         protocol = http
         host = localhost
-        port = 9570
+        port = 7022
```
AndyHWChung, line 109: don't we want it to point to the actual upscan stubs locally?

mikail-khan: That's code I've accidentally left in when testing. Will revert back.

mikail-khan: fixed

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/connectors/EmailVerificationConnector.scala
```diff
@@ -59,7 +61,11 @@ class EmailVerificationConnectorImpl @Inject() (
     val protocol = getEmailVerificationConfig[String]("protocol")
     val host     = getEmailVerificationConfig[String]("host")
     val port     = getEmailVerificationConfig[String]("port")
-    s"$protocol://$host:$port/email-verification/verification-requests"
+    val domain   = getEmailVerificationConfig[String]("domain")
+    if (domain =!= "")
```
mikail-khan, line 65: very hacky-ish - but not sure of any other clean way to achieve it

conf/application.conf
```diff
@@ -106,6 +107,7 @@ microservice {
         protocol = http
         host = localhost
         port = 9570
+        domain = cgt-property-disposals-stubs
```
AndyHWChung, line 110: shouldn't `domain` be blank if we're talking to the upscan stubs?

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/connectors/EmailVerificationConnector.scala
mikail-khan: The idea is that the domain key's value will be blank in all environments except development and staging.

AndyHWChung: to avoid the if-else in the code you could change the config values to be either `"/"` or `"/cgt-property-disposals-stubs"` - then you can just do :
```scala
 s"$protocol://$host:$port$domain/email-verification/verification-requests"
```
which might not look 100% right, but it saves you having to do this check in the code

AndyHWChung: one other option could be to make the upload submit to a test-only endpoint in our FE - that way the stub won't need to move to the public zone and we don't need to make these config changes. The stub would need to know the full url of that FE endpoint though to return in the response to the initiate call, but that's just reading something from config

mikail-khan: it will also mean running the service with test-only routes when running locally and when the QAs run the service in Jenkins

AndyHWChung: I think the test-only routes are already enabled in some of the environments - the acceptance tests require it already.

PR #683, opened 2020-07-24
CPDP-2198: [FR] Content changes for Losses and Exemptions

rambabu-posa:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/exemptionandlosses/ExemptionAndLossesController.scala
```diff
@@ -74,7 +76,23 @@ class ExemptionAndLossesController @Inject() (
     ) => Future[Result]
   ): Future[Result] =
     request.sessionData.flatMap(s => s.journeyStatus.map(s -> _)) match {
-      case Some((s, r @ FillingOutReturn(_, _, _, d, _))) =>
+      case Some((s, r @ FillingOutReturn(_, _, _, d, _))) if r.isFurtherReturn.contains(true) =>
```
mikail-khan, line 79: if you are going to call `isFurtherReturn` without the parenthesis then ideally you should remove the parenthesis from the actual function definition, so:

```scala
    def isFurtherReturn: Option[Boolean] 
```

instead of

```scala
    def isFurtherReturn(): Option[Boolean] 
```

rambabu-posa: I don't like to use parenthesis, but if someone wants use it? I feel its good to define function like this  `isFurtherReturn()`
Then its up-to user whether use of not. If we miss it in definition, then we don't have option to use parenthesis in its usage.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/exemptionandlosses/ExemptionAndLossesController.scala
```diff
@@ -269,13 +288,23 @@ class ExemptionAndLossesController @Inject() (
...
+              page =
+                if (fillingOutReturn.isFurtherReturn.contains(true))
+                  furtherReturnInYearLossesPage(
+                    _,
+                    _,
+                    fillingOutReturn.subscribedDetails.isATrust,
+                    draftReturn.representativeType(),
+                    fillingOutReturn.isFurtherReturn
+                  )
+                else
```
mikail-khan, line 300: it will hit the `else` when it is `None` - is that ok?

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/exemptionandlosses/ExemptionAndLossesController.scala
mikail-khan: the convention is if a method has side-effects include the parenthesis else if the method is not side-effecting then do not include the parenthesis. `isFurtherReturn` is not side-effecting so by convention it should have no parenthesis.

sam-miele:
LGTM - wait until Mikail signs off then merge.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/exemptionandlosses/ExemptionAndLossesController.scala
mikail-khan: see this as well @rambabu-posa https://docs.scala-lang.org/glossary/index.html#uniform-access-principle

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/exemptionandlosses/ExemptionAndLossesController.scala
rambabu-posa: yes, if its not true, means not a further return then we use existing page.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/exemptionandlosses/ExemptionAndLossesController.scala
rambabu-posa: Okay sure, thanks will pix it

rambabu-posa:
@mikail-khan yes fixed all of them

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/exemptionandlosses/further_return_in_year_losses.scala.html
```diff
@@ -0,0 +1,137 @@
...
+            fieldId = key,
+            label = label,
+            selected = selectedIndex,
+            error = form.error(key),
+            options = options,
+            helpText = None,
+            customErrorKey = Some(s"$key$userKey")
+        )
+
+        <p>@Html(messages(s"$key$furtherReturnKey$userKey.link", appConfig.furtherReturnWorkOutCGT))</p>
```
AndyHWChung, line 130: this shouldn't be read from config, it will eventually be a url you can get from reverse routing. For now, I've put "#" as the href until that page exists

PR #684, opened 2020-07-27
CPDP-2033

sam-miele:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/services/onboarding/CGTRegistrationService.scala
```diff
@@ -0,0 +1,76 @@
...
+                                  ManuallyEntered,
+                                  DerivedFromBusinessPartnerRecord
+                                )
+                              )
+    } yield subscriptionResponse
+
+    a.flatMap {
+      case SubscriptionResponse.SubscriptionSuccessful(cgtReferenceNumber) =>
+        EitherT.rightT(RepresenteeCgtReference(CgtReference(cgtReferenceNumber)))
+      case SubscriptionResponse.AlreadySubscribed                          => EitherT.leftT(Error(""))
```
AndyHWChung, line 73: a useful message would be helpful here instead of an empty string

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/services/onboarding/CGTRegistrationService.scala
```diff
@@ -0,0 +1,76 @@
...
+      subscriptionResponse <- subscriptionService.subscribe(
+                                SubscriptionDetails(
+                                  Right(representeeAnswers.name),
+                                  representeeAnswers.contactDetails.emailAddress,
+                                  representeeAnswers.contactDetails.address,
+                                  representeeAnswers.contactDetails.contactName,
+                                  SapNumber(sapNumber.body),
+                                  BusinessPartnerRecord,
+                                  ManuallyEntered,
+                                  DerivedFromBusinessPartnerRecord
```
AndyHWChung, line 65: `DerivedFromBusinessPartnerRecord` might not be true here - it's safer to put `ManuallyEntered` here for now

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/services/onboarding/CGTRegistrationService.scala
```diff
@@ -0,0 +1,76 @@
...
+
+@Singleton
+class CGTRegistrationService @Inject() (
+  val errorHandler: ErrorHandler,
+  connector: CGTPropertyDisposalsConnector,
+  subscriptionService: SubscriptionService
+)(implicit ec: ExecutionContext)
+    extends Logging {
+
+  def extractRepresenteeCgtReference(
```
AndyHWChung, line 44: `extractRepresenteeCgtReference` doesn't sound quite right - extraction implies it the cgt ref already exists. This could be something like `registerWithoutIdAndSubscribe` - that's what it's doing.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/check_all_answers.scala.html
```diff
@@ -82,6 +95,38 @@
...
 
+  @if(showSubmissionDetails) {
+      <h2 class="heading-medium">Submission Details</h2>
+      @cyaSection() {
+          @if(!isAgent) {
+              @cyaRow(
+                  messages("CGT reference number"),
+                  Html(subscribedDetails.cgtReference.value),
+                  None,
+                  "disposalMethod"
```
AndyHWChung, line 106: `disposalMethod` doesn't look right here

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/check_all_answers.scala.html
```diff
@@ -82,6 +95,38 @@
 
  @pageHeading(title)
 
+  @if(showSubmissionDetails) {
+      <h2 class="heading-medium">Submission Details</h2>
+      @cyaSection() {
+          @if(!isAgent) {
+              @cyaRow(
+                  messages("CGT reference number"),
```
AndyHWChung, line 103: it's good practice to stick these messages in the messages file - having said that, I think we're only going to need to display this page in English. Is there a way to make sure this is always in English when creating the HTML for DMS?

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/check_all_answers.scala.html
```diff
@@ -82,6 +95,38 @@
...
+                      messages("Agent reference number"),
+                      Html(messages(agentRef.value)),
+                      None,
+                      "disposalMethod"
+                  )
+              }
+          }
+          @{
+              representeeId match {
+                  case Some(a: RepresenteeNino) => cyaRow("NINO of person being represented", Html(a.value.value), None, "disposalMethod")
```
AndyHWChung, line 121:
```scala
case Some(RepresenteeNino(nino)) => .... 
```
would be cleaner so that you can do `nino.value` when you need it

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/services/onboarding/CGTRegistrationService.scala
```diff
@@ -0,0 +1,76 @@
...
+import uk.gov.hmrc.cgtpropertydisposalsfrontend.models.Error
+import uk.gov.hmrc.cgtpropertydisposalsfrontend.models.returns.RepresenteeAnswers.CompleteRepresenteeAnswers
+import uk.gov.hmrc.cgtpropertydisposalsfrontend.models.returns.RepresenteeReferenceId.RepresenteeCgtReference
+import uk.gov.hmrc.cgtpropertydisposalsfrontend.util.Logging
+import uk.gov.hmrc.http.HeaderCarrier
+
+import scala.concurrent.{ExecutionContext, Future}
+
+@Singleton
+class CGTRegistrationService @Inject() (
```
AndyHWChung, line 37: to separate the unit tests you could create a trait with the method in it as an abstract method and then implement it with an implementing class which has the logic you've got below. That way, if something else uses this service, there's less duplication in the unit tests. The classes which use this service can then be unit tested more easily.

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/CheckAllAnswersAndSubmitControllerSpec.scala
```diff
@@ -161,6 +164,18 @@ class CheckAllAnswersAndSubmitControllerSpec
       .expects(cgtReference, chargeReference, amount, returnUrl, backUrl, *, *)
       .returning(EitherT.fromEither[Future](response))
 
+  val mockRepresenteeCgtReference = RepresenteeCgtReference(CgtReference("12345"))
+
+  def mockCgtRegistrationService() =
```
AndyHWChung, line 169: for more complete coverage you could expect what inputs you're expecting and also control the response to be able to test when things go wrong

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/services/onboarding/CGTRegistrationService.scala
AndyHWChung, line 37: this class should have it's own unit test - the classes which use this service shouldn't be testing the actual logic inside the class

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/services/onboarding/CGTRegistrationService.scala
AndyHWChung: a more natural place to put this method would be in `SubscriptionService` - that already contains the methods to subscribe and register without id.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/CheckAllAnswersAndSubmitController.scala
```diff
@@ -266,7 +312,7 @@ class CheckAllAnswersAndSubmitController @Inject() (
         CompleteSingleDisposalReturn
           .fromDraftReturn(draftReturn)
           .fold[Future[Result]](
-            Redirect(routes.TaskListController.taskList())
+            Redirect((routes.TaskListController.taskList()))
```
AndyHWChung, line 315: this looks like a redundant pair of brackets here

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/check_all_answers.scala.html
```diff
@@ -83,6 +96,38 @@
...
+                      messages("checkAllAnswers.submissionDetails.agentReferencenumber"),
+                      Html(messages(agentRef.value)),
+                      None,
+                      "agentReferenceNumber"
+                  )
+              }
+          }
+          @{
+              representeeId match {
+                  case Some(RepresenteeNino(nino)) => cyaRow(messages("checkAllAnswers.submissionDetails.nino"), Html(nino.value), None, "disposalMethod")
```
AndyHWChung, line 122: the `disposalMethod` don't look right here

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/services/onboarding/SubscriptionServiceImplSpec.scala
```diff
@@ -344,6 +349,57 @@ class SubscriptionServiceImplSpec extends WordSpec with Matchers with MockFactor
...
+        representeeAnswers.contactDetails.emailAddress,
+        representeeAnswers.contactDetails.address,
+        representeeAnswers.contactDetails.contactName,
+        SapNumber(sapNumber),
+        ManuallyEnteredEmail,
+        ManuallyEnteredAddress,
+        ManuallyEnteredContactName
+      )
+
+      "user is successfully registered and subscribed returning correct cgt reference" in {
```
AndyHWChung, line 374: it'd be good to get some unit tests to cover when:
- register without id fails
- subscription fails
- subscription comes back with an "already subscribed" response

Also, this doesn't sound quite right:
```
SubscriptionServiceImpl when handing request to register without
id and subscribe must user is successfully registered and subscribed 
returning correct cgt reference
````

File: test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/CheckAllAnswersAndSubmitControllerSpec.scala [hunk 12]
```diff
@@ -161,6 +163,16 @@ class CheckAllAnswersAndSubmitControllerSpec
...
       .returning(EitherT.fromEither[Future](response))
 
+  def mockCgtRegistrationService(representeeCgtReference: RepresenteeCgtReference) =
+    (mockSubscriptionService
+      .registerWithoutIdAndSubscribe(
+        _: CompleteRepresenteeAnswers
+      )(
+        _: HeaderCarrier
+      ))
+      .expects(*, *)
```
AndyHWChung, line 173: you could still pass in `CompleteRepresenteeAnswers` to assert that the correct thing is being passed into the service and do something like `.expects(expectedAnswers, *)`. 

Also, you could pass in the response of the service like `Either[Error,RepresenteeCgtReference]` so you can do something like `.returning(EitherT.fromEither(response))` to allow you to test when things go wrong.

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/CheckAllAnswersAndSubmitControllerSpec.scala
```diff
@@ -1310,7 +1313,12 @@ class CheckAllAnswersAndSubmitControllerSpec
           inSequence {
             mockAuthWithNoRetrievals()
             mockGetSession(sessionWithJourney(completeFillingOutReturnWithRepresenteeWithNoReference))
-            mockCgtRegistrationService(mockRepresenteeCgtReference)
+            mockCgtRegistrationService(
+              Right(mockRepresenteeCgtReference),
+              mockedCompleteReturn.representeeAnswers
```
AndyHWChung, line 1318: it'd be better to factor out the `representeeAnswers` that are generated in a separate `val` so that you can reference it here instead of handling the case `None` when it can never happen. A similar thing happens in the test below

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/services/onboarding/SubscriptionServiceImplSpec.scala
```diff
@@ -371,31 +365,45 @@ class SubscriptionServiceImplSpec extends WordSpec with Matchers with MockFactor
...
 
-      "user is successfully registered and subscribed returning correct cgt reference" in {
-        val registrationResponse = Json.parse(
-          s"""
-             |{
-             |  "sapNumber" : "$sapNumber"
-             |}
-             |""".stripMargin
+      "successfully register and subscribed returning correct cgt reference" in {
+        mockRegisterWithoutId(registrationDetails)(
```
AndyHWChung, line 369: using `inSequence` like:
```scala
inSequence{
  mockRegisterWithoutId(...)
  mockSubscribe(..)
}
```
will make this test stronger as it will assert that the things happen in that order

AndyHWChung:
LGTM

PR #696, opened 2020-07-29
CPDP-2208: augment returns with clearing reason

mikail-khan:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/actions/AuthenticatedActionWithRetrievedData.scala
```diff
@@ -265,7 +265,7 @@ class AuthenticatedActionWithRetrievedData @Inject() (
           s"Agent has ${AgentsEnrolment.key} enrolment but does not have ${AgentsEnrolment.agentReferenceNumberIdentifier} identifier"
         )
       }
-      .sequence[Either[String, ?], AgentReferenceNumber]
+      .sequence[Either[String, *], AgentReferenceNumber]
```
AndyHWChung, line 268: what's the difference between `*` and `?` here?

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/finance/ClearingReason.scala
```diff
@@ -0,0 +1,51 @@
...
+  case object IncomingPayment extends ClearingReason
+  case object OutgoingPayment extends ClearingReason
+  case object WriteOff extends ClearingReason
+  case object Reversal extends ClearingReason
+  case object MassWriteOff extends ClearingReason
+  case object AutomaticClearing extends ClearingReason
+  case object SomeOtherClearingReason extends ClearingReason
+  case object InvalidClearingReason extends ClearingReason
+
+  def fromString(clearingReason: String): ClearingReason =
```
AndyHWChung, line 37: this shouldn't be needed in the FE - isn't this just to be able to read back from DES?

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/components/charge_item.scala.html
```diff
@@ -16,16 +16,50 @@
...
+            case (Some(paymentMethod), _) =>
+                    makePaymentDetailsHtml("paymentMethod", paymentMethod.toString, payment.amount.inPounds(), payment.clearingDate)
+            case (None, Some(clearingReason)) => clearingReason match {
+                case IncomingPayment | OutgoingPayment | WriteOff | Reversal | MassWriteOff | AutomaticClearing | SomeOtherClearingReason =>
+                    makePaymentDetailsHtml("clearingReason", clearingReason.toString, payment.amount.inPounds(), payment.clearingDate)
+                case InvalidClearingReason if (payment.method.isDefined) =>
+                    makePaymentDetailsHtml("paymentMethod", payment.method.fold("")(_.toString), payment.amount.inPounds(), payment.clearingDate)
+                case InvalidClearingReason if (!payment.method.isDefined) =>
+                    <li>
+                        {messages(s"viewReturn.charge.display-error")}
```
AndyHWChung, line 53: the intention here was to just display the technical error page if there was no clearing reason or payment method. We don't have content for this case.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/actions/AuthenticatedActionWithRetrievedData.scala
mikail-khan: ? is deprecated now

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/components/charge_item.scala.html
mikail-khan: It's impossible not to get both those data points. But to complete the case analysis something had to be put there.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/finance/ClearingReason.scala
mikail-khan: yeah that was copy'n'pasta - removed now

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/components/charge_item.scala.html
AndyHWChung: if we had an `Either[ClearingReason,PaymentMethod]` we could avoid handling cases which we don't expect - but that does come at the cost of backwards compatibility which wouldn't result in permanent failures in this case

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/actions/AuthenticatedActionWithRetrievedData.scala
AndyHWChung: ah ok - I'll change them over if I ever come across any as well

mikail-khan: I think I did them all across the services.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/components/charge_item.scala.html
mikail-khan: let me know you what you decide.

AndyHWChung: if we had them both as `Option` then to do what was intended we'd have to error out somewhere in the FE. If it was an `Either` then it prevents the FE from having to do anything special. However, if we ever did want to not block the user from seeing the page if we couldn't get a payment method and to just display some sort of message which is what you've done here, then having options now would make that easier in the future. It's not unfeasible that we wouldn't want to do that cos that piece of text doesn't prevent them from doing other things, so let's leave them as `Option` for now and maybe revisit this case in the future to see if we want to display some sort of message like you've done here.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/components/charge_item.scala.html
```diff
@@ -44,11 +44,11 @@
...
                     makePaymentDetailsHtml("paymentMethod", paymentMethod.toString, payment.amount.inPounds(), payment.clearingDate)
             case (None, Some(clearingReason)) => clearingReason match {
-                case IncomingPayment | OutgoingPayment | WriteOff | Reversal | MassWriteOff | AutomaticClearing | SomeOtherClearingReason =>
+                case IncomingPayment | OutgoingPayment | WriteOff | Reversal | MassWriteOff | AutomaticClearing =>
                     makePaymentDetailsHtml("clearingReason", clearingReason.toString, payment.amount.inPounds(), payment.clearingDate)
-                case InvalidClearingReason if (payment.method.isDefined) =>
+                case SomeOtherClearingReason if (payment.method.isDefined) =>
                     makePaymentDetailsHtml("paymentMethod", payment.method.fold("")(_.toString), payment.amount.inPounds(), payment.clearingDate)
-                case InvalidClearingReason if (!payment.method.isDefined) =>
+                case SomeOtherClearingReason if (!payment.method.isDefined) =>
```
AndyHWChung, line 51: do you even need this case now? Couldn't you just have:
```scala
case (Some(paymentMethod), _) => ...
case (_, Some(clearingReason)) =>
  makePaymentDetailsHtml("clearingReason", clearingReason.toString, payment.amount.inPounds()
```
the order of the pattern match means you don't need to check if `paymentMethod` is defined or not after that first case

mikail-khan: looking at it again, the logic is now incorrect - let me fix it

AndyHWChung:
LGTM

PR #710, opened 2020-08-05
CPDP-2623: changed opening times for Welsh

nubz:


conf/messages.cy
```diff
@@ -327,9 +327,7 @@ iv.insufficientEvidence.textphone.title=Ffôn testun:
 iv.insufficientEvidence.textphone.number=0300 200 3319
 iv.insufficientEvidence.outside-uk-telephone.title=Y tu allan i’r DU:
 iv.insufficientEvidence.outside-uk-telephone.number=+44 300 200 1900
-iv.insufficientEvidence.openingTimes=Oriau agor:
-iv.insufficientEvidence.openingTimes.p1=Dydd Llun – Dydd Gwener: 08:00 – 20:00
-iv.insufficientEvidence.openingTimes.p2=Dydd Sadwrn: 08:00 – 16:00
+iv.insufficientEvidence.openingTimes=<p>Oriau agor:</p><p>Dydd Llun – Dydd Gwener: 08:00 – 16:00</p>
```
AndyHWChung, line 330: why are the opening times different in Welsh? This is Mon-Fri 8am to 4pm but in English it's Mon-Fri 8am to 8pm and Sat 8pm to 4pm

nubz: If you look at the ticket - it's because these are the times for the Welsh speaking line

nubz: wasn't pointed out in original translation - so I had corrected it and now they are explaining that it was right..

nubz: this is why I had to put the HTML into the message rather than try and work out language in the template...

AndyHWChung:
LGTM

PR #711, opened 2020-08-05
CPDP-2206: fix date logic

mikail-khan:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/confirmation_of_submission.scala.html
```diff
@@ -112,7 +112,7 @@
   case _ => s"$key$userKey.printPage"
   }}
 
-@isLateReturn = @{completionDate.value.plusDays(30L).isAfter(TimeUtils.today)}
+@isLateReturn = @{completionDate.value.plusDays(30L).isBefore(TimeUtils.today)}
```
AndyHWChung, line 115: oh yeah, this makes more sense. If today is 1st August and the completion date was 1st July then
```
isLate 
  = completionDate.value.plusDays(30L).isBefore(TimeUtils.today)
  = (31st July).isBefore(1st August)
  = true  
```
which is correct since the calculated due date (31st July) is before today (1st August). `isAfter` would have given the incorrect result there. That's right isn't it?

AndyHWChung:
LGTM

PR #717, opened 2020-08-06
CPDP-2274 qa updates

sam-miele:
Fixes:

Remove the further help page from the "check your answers" check and explicitly link there were needed to avoid loop in Multiple Disposals

Only show return to summary when complete

Back link for "How many properties"

### DIFF
--- a/app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/triage/CommonTriageQuestionsController.scala
+++ b/app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/triage/CommonTriageQuestionsController.scala
@@ -126,24 +126,30 @@ class CommonTriageQuestionsController @Inject() (
                   )
                 ),
               { individualUserType =>
-                val answers    = triageAnswersFomState(state)
-                val redirectTo =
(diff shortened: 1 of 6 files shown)


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/triage/CommonTriageQuestionsController.scala
```diff
@@ -122,7 +127,10 @@ class CommonTriageQuestionsController @Inject() (
                 ),
               { individualUserType =>
                 val answers    = triageAnswersFomState(state)
-                val redirectTo = redirectToCheckYourAnswers(state)
+                val redirectTo =
+                  if (state.fold(_.isFurtherReturn, _.isFurtherReturn).contains(true))
```
AndyHWChung, line 131: you'd have to do `isFurtherReturn` on an updated state with the updated answers in it - `isFurtherReturn` depends on the `IndividualUserType` which has just been submitted in this endpoint. `state` won't always contain the answer that was just submitted

AndyHWChung:
LGTM

PR #718, opened 2020-08-06
CPDP-2466: add isFutherReturn flag

mikail-khan:


mikail-khan:
@AndyHWChung can you re-review to make sure we are on the same page

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/returns/CompleteReturn.scala
```diff
@@ -81,7 +82,8 @@ object CompleteReturn {
               u,
               maybeCompleteRepresenteeAnswers,
               gainOrLossAfterReliefs,
-              hasAttachments = true
+              hasAttachments = true,
+              isFirstReturn = true
```
AndyHWChung, line 86: a draft return is not always necessarily a first return so this value is incorrect. I originally suggested sticking this `isFirstReturn` flag in the `ViewingReturn` journey status to avoid having to set it here. While we're filling our a return, that information is already contained within `FillingOutReturn`, so there'd be a duplication of information sticking it here.

mikail-khan: If see the BE draft PR, the flag is set there in CompleteReturn  and what I did was to map that type for type in the FE. Is that not correct? Because I did that the problem above occurs. What I think you are saying now is that once the response is returned from the BE, determine the flag and then set it in the `ViewingReturn` and not in the `CompleteReturn` subtypes?

AndyHWChung: ah ok, I can see it being set in the BE now when reading back the return from DES. Putting the flag in `CompleteReturn` though means that you also need to send it in the submit return request from the FE to the BE which is why you're having to populate the flag here in the `apply` methods that take you from a draft return to a complete return. Even though it's not being read by the BE in the submit return request it's still there together with the actual `isFurtherReturn: Boolean` parameter you've put in the submit return request. Sticking it outside means that you can avoid having to do that but it means you'll have to update the response to get a return to include the `isFirstReturn` flag outside of `CompleteReturn` which would mean updating the JSON structure of the response.

AndyHWChung: the way forward with the least amount of changes then would be to add in an extra parameter to these apply methods to give the proper value of `isFirstReturn`, which would be equal to `!isFurtherReturn` which should be available from the place they are being called.

AndyHWChung: it seems cleaner to stick it outside though so you don't have to do the extra work of populating it when you don't need it

mikail-khan: I think what you're saying now is that instead of augmenting `CompleteReturn` in the BE with the flag, return a new type which is a composite of the `CompleteReturn` and the flag. That way the subtypes of `CompleteReturn` don't need to be changed in the BE and neither in the FE.

AndyHWChung: yeah, that's basically it just in a lot fewer words :smile:

mikail-khan: added a new type called `DisplayReturn` - can you re-review the FE and BE draft PRs. If fine, then I will fix the FE broken tests.

in particular can you check the changes around the new `StartingAmendReturn` type you added yesterday

mikail-khan: @AndyHWChung :point_up:

AndyHWChung: changes look good - it's good to have the new field somewhere in `StartingAmendReturn` - we're gonna need that flag for something very soon

mikail-khan: added flag to `StartingAmendReturn`

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/ViewReturnController.scala
```diff
@@ -73,7 +74,7 @@ class ViewReturnController @Inject() (
               subscribedDetails,
               sentReturn.representativeType,
               sentReturn.isIndirectDisposal,
-              None
+              Some(isFirstReturn)
```
AndyHWChung, line 77: shouldn't this be `Some(!isFirstReturn)` - the template is expecting `isFurtherReturn`

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/view_return.scala.html
```diff
@@ -71,6 +73,67 @@
 @title = @{messages("viewReturn.title")}
 @amountOwed = @{returnSummary.mainReturnChargeAmount}
 
+@glarAmount = @{
```
AndyHWChung, line 76: it would take up less space if you just extracted out the glar value where you need it rather than doing a big patter match here - where you need it, a particular type of complete return will be in scope for you to extract it out of

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/view_return.scala.html
```diff
@@ -179,6 +242,11 @@ <h2 class="heading-medium">@messages("task-list.acquisition-details.link")</h2>
           <h2 class="heading-medium">@messages("task-list.relief-details.link")</h2>
           @reliefDetailsSummary(s.reliefDetails, subscribedDetails.isATrust, representativeType)
 
+          @if(glarAmount.isDefined && isFurtherReturn.contains(true)){
```
AndyHWChung, line 245: e.g. here you could just do:
```scala
s.gainOrLossAfterReliefs.map{ g =>
  ...
}
```
then you wouldn't have to do `.getOrElse`

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/ViewReturnController.scala
mikail-khan: actually, it should be fixed in the BE - I think I did have that at one point but when I changed the implementation y'day to what we discussed I forgot to add it in again. Fixed in BE now.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/view_return.scala.html
mikail-khan: fixed

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/view_return.scala.html
mikail-khan: fixed

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/ViewReturnController.scala
AndyHWChung: the template is expecting `isFurtherReturn` though so shouldn't it still be `!isFirstReturn` here?

mikail-khan: if you look at what you wrote in CPDP-2473:
> return flag to FE to say whether or not a first return is being returned
I flip the boolean value to reflect whether it is a first return or not.

If you think that this should negated here then it really doesn't make sense to have a field called `isFirstReturn` - it should just be `isFurtherReturn`.

AndyHWChung: it's more accurate to return `isFirstReturn` from the BE when we display a return since when we do amends, the value of that flag will be the same for further returns and amended returns. We're treating amended returns in a similar way to further returns but they're not the same thing.

mikail-khan: OK. I'll flip the logic in FE. Very confusing design :man_shrugging:

AndyHWChung:
LGTM

PR #731, opened 2020-08-13
CPDP-2199: Further returns guidance page

rambabu-posa:


test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/triage/FurtherReturnGuidanceControllerSpec.scala
```diff
@@ -0,0 +1,183 @@
...
+              )._1
+            )
+          }
+
+          checkIsTechnicalErrorPage(performAction("abc"))
+        }
+
+      }
+
+      "display the page" when {
```
AndyHWChung, line 134: it'd be good to unit test the different h1's for the different user types as well - only individual "self" users are being tested atm

AndyHWChung: making sure the different journey statuses are covered would also be good (`StartingNewDraftReturn`, `FillingOurReturn` and `StartingAmendReturn`). It'd be enough to make sure the tests for each user type covers all the different journey statuses that are allowed. we don't necessarily need every user type with all journey status types

rambabu-posa: Is it okay how we did for other controllers like `ReliefDetailsControllerSpec`?

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/triage/FurtherReturnGuidanceController.scala
```diff
@@ -0,0 +1,134 @@
...
+          _ => None,
+          _.newReturnTriageAnswers.fold(
+            _.representativeType(),
+            _.representativeType()
+          )
+        ),
+        _.draftReturn.representativeType()
+      )
+
+  private def withState(request: RequestWithSessionData[_])(
```
mikail-khan, line 90: could you rename this to `withJourneyState` as that makes it more clear what state it is referring to

PR #732, opened 2020-08-14
CPDP-2710 net gain guidance

nubz:


test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/triage/FurtherReturnGuidanceControllerSpec.scala
```diff
@@ -651,6 +651,66 @@ class FurtherReturnGuidanceControllerSpec
...
+
+          def test(
+            individualUserType: IndividualUserType,
+            userType: UserType
+          ): Unit = {
+            inSequence {
+              mockAuthWithNoRetrievals()
+              mockGetSession(
+                sessionDataWithFillingOutReturnForSingleDisposals(
+                  IncompleteSingleDisposalTriageAnswers.empty.copy(
```
rambabu-posa, line 676: Please add unit tests for Multiple Disposals too and also cover other journey like "StartingNewDraftReturn and StartingAmendReturn"

nubz: problem is startingNewDraftReturn and startingAmendReturn is those states only apply to returns that have not yet been saved as a draft - it is only possible to get to the net gain page in section 3 whilst filling out a saved draft return (after triage is complete then draft is saved and other sections can be accessed) - also the net gain page and guidance page are the same with no variations for single or multiple - but I can add the same test for multiple answers anyway just to be sure...

nubz: added multiple case

rambabu-posa:
LGTM

PR #735, opened 2020-08-17
CPDP-2199: removed caption padding, changed heading classes in exampl…

nubz:
…es and stopped button push effect


AndyHWChung:
LGTM

PR #745, opened 2020-08-18
Cpdp 2378 fixes 3

sam-miele:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/gainorlossafterreliefs/GainOrLossAfterReliefsController.scala
```diff
@@ -102,13 +102,42 @@ class GainOrLossAfterReliefsController @Inject() (
...
                   val updatedAmount      = AmountInPence.fromPounds(value)
                   val updatedDraftReturn =
-                    draftReturn.fold(
-                      _.copy(gainOrLossAfterReliefs = Some(updatedAmount)),
-                      _.copy(gainOrLossAfterReliefs = Some(updatedAmount)),
-                      _.copy(gainOrLossAfterReliefs = Some(updatedAmount)),
-                      _.copy(gainOrLossAfterReliefs = Some(updatedAmount)),
-                      _.copy(gainOrLossAfterReliefs = Some(updatedAmount))
-                    )
+                    if (fillingOutReturn.isFurtherReturn.contains(true))
```
AndyHWChung, line 105: gain or loss after reliefs is always for a further return so you don't need to do this check here - it'll always be true

AndyHWChung:
LGTM

PR #749, opened 2020-08-18
CPDP-2695: amend cya to return

AndyHWChung:
amend treated as further return; prevent saving draft return on amend


AndyHWChung:
main updates are:
- some controllers now convert from `StartingToAmendReturn` to `FillingOutReturn` with the help of a new `StartingToAmendToFillingOutReturnBehaviour` trait. This takes care of wiping out section 3
- in most places where `isFurtherReturn` was used, it is now `isFurtherReturnOrAmendReturn` as amend returns are to be treated like further returns by default
- calls to save a draft return will do nothing for an amend return
- link to save and come back later is not functional for an amend return

rambabu-posa (approved):
LGTM

PR #753, opened 2020-08-19
CPDP-2607: unmet dependencies

AndyHWChung:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/components/submit_button.scala.html
```diff
@@ -14,9 +14,12 @@
  * limitations under the License.
  *@
 
+@import play.twirl.api.Html
+
 @this()
 
-@(label: String)
+@(label: String, otherHtml: Option[Html] = None)
```
AndyHWChung, line 21: @nubz  I don't think this is the prettiest thing but it allowed me to stick in a cancel button where I needed it - is there a better way of doing this do you think?

nubz: seems a lot for one use case, but who knows we might need it for something else...

PR #759, opened 2020-08-21
CPDP-2588: [AR] DisposalDate exit page for different taxyear

rambabu-posa:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/triage/MultipleDisposalsTriageController.scala
```diff
@@ -927,24 +928,72 @@ class MultipleDisposalsTriageController @Inject() (
...
+                                           updatedAnswers,
+                                           d =>
+                                             d.bimap(
+                                               multipleIndirect =>
+                                                 multipleIndirect.copy(
+                                                   exampleCompanyDetailsAnswers = multipleIndirect.exampleCompanyDetailsAnswers,
+                                                   yearToDateLiabilityAnswers = None,
+                                                   gainOrLossAfterReliefs = None
+                                                 ),
+                                               multiple =>
```
AndyHWChung, line 968: you shouldn't be in a non-indirect disposal case in this endpoint since we're talking about the disposal date of shares

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/triage/MultipleDisposalsTriageController.scala
```diff
@@ -927,24 +928,72 @@ class MultipleDisposalsTriageController @Inject() (
...
+                        logger.warn("Could not find tax year or update session", e)
+                        errorHandler.errorResult()
+                      },
+                      existingTaxYear =>
+                        if (!isValidTaxYear(existingTaxYear, shareDisposalDate.value))
+                          Redirect(routes.CommonTriageQuestionsController.amendReturnDisposalDateDifferentTaxYear())
+                        else
+                          Redirect(routes.MultipleDisposalsTriageController.checkYourAnswers())
+                    )
+                  case _                                                                    =>
```
AndyHWChung, line 995: I'm not sure that you need two cases in the pattern match when it's an amend return and when it's not - they'll both look very similar. The logic with the existing tax year is still wrong as well. Everything could be the same until the right of the fold of the handling of the result - all that needs to go in there is something like:
```scala
if(state.exists(_.originalReturn.disposalDate =!= taxYear))) Redirect(newExitPage)
else Redirect(cya)
```
The original return data for an amend return is in `FillingOutReturn` under a new field  in it called `originalReturn` - that field is defined only if it's an amend return
`.disposalDate` doesn't exist above, but you might find it useful to create it cos you'll need it in multiple places

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/triage/CommonTriageQuestionsController.scala
```diff
@@ -396,6 +398,25 @@ class CommonTriageQuestionsController @Inject() (
       }
     }
 
+  private def amendReturnDisposalDateBackLink(state: Either[StartingNewDraftReturn, FillingOutReturn]): Call =
+    state.fold(
+      _ => controllers.routes.StartController.start(),
+      _.draftReturn.fold(
+        _ => routes.MultipleDisposalsTriageController.whenWereContractsExchanged(),
```
AndyHWChung, line 405: this is the endpoint for completion date for multiple disposals not disposal date. The case for both multiple and multiple mixed use should come from the disposal date question in the example property details section

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/triage/CommonTriageQuestionsController.scala
```diff
@@ -396,6 +398,25 @@ class CommonTriageQuestionsController @Inject() (
...
+        _ => routes.SingleDisposalsTriageController.whenWasDisposalDate(),
+        _ => routes.SingleDisposalsTriageController.disposalDateOfShares(),
+        _ => routes.MultipleDisposalsTriageController.disposalDateOfShares(),
+        _ => controllers.routes.StartController.start()
```
AndyHWChung, line 409: single mixed use should come from the disposal date question in `MixedUsePropertyDetailsController.scala`

rambabu-posa: As I wan unable to find any scenario for SingleMixedUse in CPDP-2588 story, I used this one 
`_ => controllers.routes.StartController.start()`

And also `MixedUsePropertyDetailsController` does not have any disposal date question. 
Please suggest me on this.

AndyHWChung: oh yeah, sorry - not sure what I was looking at. Looks like for single mixed use it uses the same disposal date question in the triage section, so it should go back to there in the last case here

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/triage/MultipleDisposalsTriageControllerSpec.scala
```diff
@@ -3710,6 +3711,7 @@ class MultipleDisposalsTriageControllerSpec
         }
 
       }
+
```
AndyHWChung, line 3714: it'd be good to get some unit tests for the new behaviour in the `MultipleDisposalstriageController`

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/triage/SingleDisposalsTriageControllerSpec.scala
```diff
@@ -2379,10 +2387,11 @@ class SingleDisposalsTriageControllerSpec
               ),
               checkIsRedirect(
                 _,
-                routes.SingleDisposalsTriageController.checkYourAnswers()
+                routes.CommonTriageQuestionsController.amendReturnDisposalDateDifferentTaxYear()
               ),
               () => mockGetTaxYear(newDisposalDate.value)(Right(Some(taxYear))),
-              Some(sample[CompleteRepresenteeAnswers].copy(dateOfDeath = Some(DateOfDeath(today.minusDays(1L)))))
+              Some(sample[CompleteRepresenteeAnswers].copy(dateOfDeath = Some(DateOfDeath(today.minusDays(1L))))),
+              isAmend = true
```
AndyHWChung, line 2394: are there still any unit tests which test it when it isn't an amend here?

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/triage/CommonTriageQuestionsControllerSpec.scala
```diff
@@ -154,36 +157,110 @@ class CommonTriageQuestionsControllerSpec
       draftReturn = draftReturn,
       subscribedDetails = sample[SubscribedDetails].copy(name = name),
       previousSentReturns = previousReturns,
-      amendReturnData = if (isAmend) Some(sample[AmendReturnData]) else None
+      amendReturnData =
+        if (isAmend)
+          Some(
+            sample[AmendReturnData].copy(
+              shouldDisplayGainOrLossAfterReliefs = true
```
AndyHWChung, line 164: setting the flag to `true` here prevents the tests from testing the `false` to `true` behaviour when it is an amend. I'd leave this as it was and use the `withForceDisplayGainOrLossAfterReliefsForAmends` on `FillingOutReturn` where you need it in your tests

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/triage/CommonTriageQuestionsControllerSpec.scala
```diff
@@ -2270,6 +2347,139 @@ class CommonTriageQuestionsControllerSpec
...
+                )._1
+              )
+
+              test(
+                routes.SingleDisposalsTriageController.disposalDateOfShares()
+              )
+            }
+          }
+
+          "the user is on a multiple indirect disposals journey" in {
```
AndyHWChung, line 2452: there's a case for single mixed use as well which you could add

test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/triage/SingleDisposalsTriageControllerSpec.scala
```diff
@@ -5794,15 +5809,23 @@ class SingleDisposalsTriageControllerSpec
...
-        draftReturn = draftReturn,
-        amendReturnData = if (isAmend) Some(sample[AmendReturnData]) else None
-      )
+    val fillingOutReturn = sample[FillingOutReturn].copy(
+      draftReturn = draftReturn,
+      amendReturnData =
+        if (isAmend)
+          Some(
+            sample[AmendReturnData].copy(
+              shouldDisplayGainOrLossAfterReliefs = true
```
AndyHWChung, line 5824: this also prevents testing of the `false to `true` behaviour - I'd leave it as it was and make sure the `updateJourney` you supply to the function takes care of setting that flag correctly

AndyHWChung:
LGTM

PR #775, opened 2020-08-26
Cpdp 2751

sam-miele:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/util/WelshSorting.scala
```diff
@@ -0,0 +1,106 @@
...
+          divergentIndex = divergentIndex + 1
+
+        //Is a substring, and return shorter
+        if (divergentIndex === maxLength)
+          compareListOfWeights.length > listOfWeights.length
+        else
+          listOfWeights(divergentIndex) < compareListOfWeights(divergentIndex)
+      }
+
+    def clean(code: String): String =
```
AndyHWChung, line 87: this could be written more simply as:
```scala
StringUtils.stripAccents(code.toLowerCase().stripPrefix(YR).filter(_.isAlphaChar))
```

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/util/WelshSorting.scala
```diff
@@ -0,0 +1,106 @@
...
+        val truncated = code.substring(2, code.length)
+        StringUtils.stripAccents(truncated).replaceAll("[^A-Za-z0-9] ", "").toLowerCase()
+      } else StringUtils.stripAccents(code).replaceAll("[^A-Za-z0-9] ", "").toLowerCase()
+
+    @scala.annotation.tailrec
+    def convertToCharacterValues(str: String, acc: List[Int]): List[Int] =
+      str match {
+        case ""                                        => acc
+        case s if validMultiCharacterWelshCharacter(s) =>
+          convertToCharacterValues(s.substring(2), acc ::: List(WELSH_ALPHABET(s.substring(0, 2))))
```
AndyHWChung, line 99: the case where something doesn't exist in the map should be handled explicitly - at the moment it'll throw an exception

sam-miele: It should be impossible for a character to not exist in the map due to the cleaning earlier.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/util/WelshSorting.scala
```diff
@@ -0,0 +1,101 @@
...
+        case s                                         => convertToCharacterValues(s.substring(1), acc ::: List(WELSH_ALPHABET(s.substring(0, 1))))
+      }
+
+    @scala.annotation.tailrec
+    def f(l1: List[Int], l2: List[Int]): Boolean =
```
AndyHWChung, line 90: I think there must be better name than `f`  - also `f`, `convertToCharacterValues`, `clean` and `validMultiCharacterWelshCharacter` shouldn't be. I'd make them private or stick them inside  `isBeforeInWelsh`

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/util/WelshSorting.scala
```diff
@@ -0,0 +1,101 @@
...
+      (l1 -> l2) match {
+        case (h1 :: Nil, h2 :: Nil) => h1 < h2
+        case (h1 :: Nil, h2 :: _)   => if (h1 === h2) true else h1 < h2
```
AndyHWChung, line 93: here this is just equivalent to `h1 <= h2`

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/util/WelshSorting.scala
```diff
@@ -0,0 +1,101 @@
...
+        case (h1 :: _, h2 :: Nil)   => if (h1 === h2) false else h1 < h2
```
AndyHWChung, line 94: and this is equivalent to `h1 < h2`

sam-miele: Not 100%. In the case that we have a substring, we want to assert that the substring comes first

i.e

("aa", "a") -> ("a", "aa") and...
("a", "aa") -> ("aa", "a")

So for this case we explicitly return false if they're equal, as h1 <= h2 would return true for that case.

AndyHWChung: so with:
```scala
if (h1 === h2) false else h1 < h2
```
if `h1 = h2 = 1` then the statement returns `false`, and `h1 < h2` is equivalent to `1 < 1` which will also return `false`, so the first check for equality is redundant isn't it?

sam-miele: Hah! You're right. We'll do `<=` for the first case and `<` for the second. That'll be fine.

AndyHWChung:
LGTM

PR #783, opened 2020-09-01
CPDP-2902: allow zero ytd

AndyHWChung:


rambabu-posa (approved):
Yes error messages are saying 0 or more.
LGTM

PR #791, opened 2020-09-03
CPDP-2826: Further return flag cleanup

rambabu-posa:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/landing_page.scala.html
```diff
@@ -173,9 +173,7 @@ <h2 class="heading-medium" id="capacitorOrPersonalRepresentative">@messages("lan
 
     <p>@Html(messages("landingPage.capacitorOrPersonalRepresentative.p5"))</p>
     <p>@Html(messages("landingPage.capacitorOrPersonalRepresentative.p6"))</p>
-    @if(!appConfig.furtherReturnsEnabled) {
-        <p>@Html(messages("landingPage.capacitorOrPersonalRepresentative.p7", appConfig.contactHmrc))</p>
```
AndyHWChung, line 177: we should be able to remove this message key from the messages file if it's not being used anymore

rambabu-posa: yes good spot

PR #793, opened 2020-09-03
CPDP-2841

sam-miele:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/triage/CommonTriageQuestionsController.scala
```diff
@@ -145,7 +146,9 @@ class CommonTriageQuestionsController @Inject() (
                   updateIndividualUserType(state, individualUserType)
 
                 val redirectTo =
-                  if (
+                  if (individualUserAmendingRepresenteeFromSelf(state, oldIndividualUserType, individualUserType))
```
AndyHWChung, line 149: you don't need something as elaborate as this I don't think. Isn't it just enough to go to this branch if 
```scala
state.fold(_ => false, _.isAmendReturn) && (individualUserType =!= Self)
```
i.e. it's an amend return and they didn't choose `Self`?

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/triage/CommonTriageQuestionsController.scala
```diff
@@ -434,6 +441,11 @@ class CommonTriageQuestionsController @Inject() (
       }
     }
 
+  def amendWhoAreYouSubmittingFor(): Action[AnyContent] =
```
AndyHWChung, line 444: getting a small unit test to check things like the title and the back link location of this page would be good

PR #811, opened 2020-09-10
CPDP-2755: Check Tax Due backlink fix

rambabu-posa:


test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/yeartodatelliability/YearToDateLiabilityControllerSpec.scala
```diff
@@ -4982,7 +4982,8 @@ class YearToDateLiabilityControllerSpec
             yearToDateLiability: AmountInPence,
             userType: UserType,
             individualUserType: IndividualUserType,
-            expectedP1Key: String
+            expectedP1Key: String,
+            expectedBackLink: Call
```
AndyHWChung, line 4986: the back link location is controlled by what's in the method rather than what's outside of it since this method is always generating complete answers, so it doesn't make sense to have it as an input parameter, unless you make the current ytd answers part of the input as well. If you did that, you'd be able to test the back link location when the answers are not complete

rambabu-posa: Can I remove that logic?

AndyHWChung: the way to get the fullest test coverage would be to have both the current ytd answers and the expected back link as parameters to this function so that you can control both at the point of calling it. Then you can test both scenarios: when the answers are incomplete and when they are complete.

AndyHWChung:
LGTM

PR #814, opened 2020-09-11
CPDP-2961: YTD Cya for Period of Admin.

rambabu-posa:


AndyHWChung:
the spec need to be updated here to account for period of admin as well:
https://github.com/hmrc/cgt-property-disposals-frontend/blob/master/test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/yeartodatelliability/YearToDateLiabilityControllerSpec.scala#L7675

AndyHWChung:
LGTM

PR #835, opened 2020-09-18
CPDP-2825: updated README

AndyHWChung:


mikail-khan:
I don't think you need all this. You just typically write how to launch the service and what service manager profile you need to run. This is all too much and not necessary.

AndyHWChung:
it's better than not having it - having something to give context to what the code is about is useful to have.

I'll put in the stuff about service manager and running it though

mikail-khan:
service design is typically put in the confluence space for the project - but if you want to put it all here fill your boots.

AndyHWChung:
when we come to the GDS live assessment, instructions on how to run the service with service manager isn't going to make much sense to them - at least they'll be able to make sense of the other stuff.
My aim was that you'd still be able to follow what's happening inside the code if you only had access to what's public.

mikail-khan:
For a GDS assessment, the technical assessment will be done by a person familiar with the MDTP microservice architecture. It will totally make sense to that person (this is my experience of doing various assessments). 

Another the problem with putting it here and then transferring the service is that when the design/code changes then this monologue has to be updated and if not then it goes out of sync which will lead to confusion, especially since developers rarely update such long READMEs - upscan-initiate being a case in point.

Like I said, if you want to put all this in then go ahead but this is not the right place for it from experience and practice.

AndyHWChung:
for internal assessment, it's true that they'll be familiar with MDTP - but the last GDS assessment I did they weren't.

I took care with what I put in there that it described things which weren't likely to ever change. It's light in specifics and only goes as deep as it needs to.

PR #841, opened 2020-09-22
CPDP-2800 confirmation page for amends

nubz:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/confirmation_of_submission.scala.html
```diff
@@ -135,7 +136,7 @@
     <h1 class="heading-xlarge">@title</h1>
     @if(!isPeriodOfAdmin) {
       <span class="heading-large" style="font-weight: normal;">
-        @messages(s"$key.taxOwed")
+        @messages(s"$key.${if(isAmend) "revisedLiability" else "taxOwed"}")
```
rambabu-posa, line 139: I think its good to extract this logic into a variable like
`@x = @{ if(isAmend) "revisedLiability" else "taxOwed" }`
then
@messages(s"$key.$x")

nubz: ok, that's done - better to have less logic in key names!

rambabu-posa (approved):
LGTM

PR #847, opened 2020-09-24
CPDP-2801: content changes for view amends

mikail-khan:


mikail-khan:
@AndyHWChung & @rambabu-posa please could you review these changes so far. I attempted to make the content changes for the deltas but the code got ugly quickly which makes me doubt these changes are optimal. If these changes are ok then another attempt can be made on the delta changes.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/view_return.scala.html
```diff
@@ -79,7 +92,7 @@
     @cyaSection() {
         @cyaRow(
             messages(s"$key.sentToHmrc"),
-            Html(TimeUtils.govDisplayFormat(returnSummary.submissionDate)),
+            if(returnSummary.isRecentlyAmended){amendDate}else{Html(TimeUtils.govDisplayFormat(returnSummary.submissionDate))},
```
AndyHWChung, line 95: you should be able to do something like: 
```scala
Html(TimeUtils.govDisplayFormat(returnSummary.lastyUpdatedDate.getOrElse(returnSummary.submissionDate)))
```
to avoid having to pattern match with `amendDate` up there with the `None` case

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/components/return_charges.scala.html
```diff
@@ -15,49 +15,46 @@
...
+            <tr style="vertical-align:bottom;">
+                <th scope="col">@messages("viewReturn.charges.description")</th>
+                <th scope="col">@messages("viewReturn.charges.payBy")</th>
+                <th scope="col">@messages("viewReturn.charges.amount")</th>
+                <th scope="col">@messages("viewReturn.charges.due")</th>
+                <th scope="col">@messages("viewReturn.charges.status")</th>
+            </tr>
+            </thead>
+            <tbody id="returnPaymentDetails-@returnSummary.submissionId">
+            @{ returnSummary.charges.map(chargeItem(_))}
```
AndyHWChung, line 57: you'll need to include the delta charge here as well if it exists. The description for it should be `Updated tax liability` if it exists.

AndyHWChung: but also, if there is no delta charge the description of the main return charge type needs to be `Updated tax liability` as well for all amend returns

mikail-khan: Yeah, that's where I was passing in the delta object through but then ripped it all out. So looks like I was on the right lines there. Do I have to match the charge reference in the delta object with the charge reference on the Charge object and update the charge description?

AndyHWChung: yeah, those are the two rows which should appear. The charge type's which need their messages updating for amend returns are `UkResidentReturn` and `NonUkResidentReturn`. In the message files we've currently got:
```
viewReturn.chargeType.UkResidentReturn=Tax payment
viewReturn.chargeType.NonUkResidentReturn=Tax payment
```
so you could just add some extra ones there like:
```
viewReturn.amend.chargeType.UkResidentReturn=Updated tax liability
viewReturn.amend.chargeType.NonUkResidentReturn=Updated tax liability
```
and make the template use those keys when needed. Then you could stick the delta charge row on the bottom

mikail-khan: can I also assume that given there is a delta then an amend has been done - ie I don't need to explicitly pass in a flag to the charge_item template to implement the logic as per the ticket; that's what I was doing before

AndyHWChung: a delta charge implies an amend return has been done, but an amend return does not imply a delta charge will be raised. So you may need to pass in a separate flag cos the content needs to apply for all amend returns

mikail-khan:
> you need to move the delta charge out of `chargeItem` actually - what's gonna happen atm if there is more than one charge (e.g. the main return charge and a late filing penalty) is that you're gonna get the delta charge showing up more than once. You could stick the delta charge row to the table manually here unless you can convert the `ReturnCharge` to a `Charge` somehow and give it the right description still.

The content of the table is not totally done yet - but what I am interested in is if that `if-else` ladder looks correct?

AndyHWChung:
> > you need to move the delta charge out of `chargeItem` actually - what's gonna happen atm if there is more than one charge (e.g. the main return charge and a late filing penalty) is that you're gonna get the delta charge showing up more than once. You could stick the delta charge row to the table manually here unless you can convert the `ReturnCharge` to a `Charge` somehow and give it the right description still.
> 
> The content of the table is not totally done yet - but what I am interested in is if that `if-else` ladder looks correct?
 
You might not need such a big if-else in `chargeItem` - it appears to be only the charge description label which changes to have an extra `amend` in the message key. If you introduce the `amend` part in the message key, all of the charge descriptions will need to have that key defined in the messages even if they don't change for amends.

The part about the delta charge needs to come out of `chargeItem` and should be appended to the table so that it only appears once.

mikail-khan:
I ripped out the if-else ladder and re-worked the return-charges template. Also I added a helper method to convert a ReturnCharge to a Charge so that the delta charges can be pushed through the charges template. Still not sure about how the delta charges should be displayed. I made another attempt but reversed out the code but left what I think should be structure of the boolean logic to display this when it is an amend. Hoping you can pick it up from there.

PR #859, opened 2020-09-30
CPDP-2801 only using positive charges

nubz:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/JourneyStatus.scala
```diff
@@ -81,7 +81,7 @@ object JourneyStatus {
   object Subscribed {
     implicit class SubscribedOps(private val s: Subscribed) extends AnyVal {
       def totalLeftToPay(): AmountInPence =
-        AmountInPence(s.sentReturns.map(_.totalOutstanding.value).sum)
+        AmountInPence(s.sentReturns.map(_.totalOutstanding.withFloorZero.value).sum)
```
AndyHWChung, line 84: would `totalOutstanding` ever be below zero? It's defined like this:
```scala
  implicit class ReturnSummaryOps(private val r: ReturnSummary) extends AnyVal {

    def totalCharges(): AmountInPence =
      AmountInPence(r.charges.map(_.amount.withFloorZero.value).sum)

    def totalPaid(): AmountInPence =
      AmountInPence(r.charges.map(_.totalPaid().value).sum)

    def totalOutstanding(): AmountInPence = totalCharges() -- totalPaid()

  }
```
so `totalCharges` is always greater than or equal to zero. Should we be flooring the `totalPaid` instead?

nubz: yes can swap it into there instead

AndyHWChung:
LGTM

PR #900, opened 2020-10-22
omitting income and allowance answers for trusts

nubz:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/ytdliability/non_calculated_summary.scala.html
```diff
@@ -112,24 +112,27 @@
         )
     }
 
+   @if(!isATrust) {
```
AndyHWChung, line 115: was this actually showing for trusts? It should have been that those answers would be `None` for trusts so that it wouldn't show

PR #902, opened 2020-10-22
CPDP-3155: remove you need to calculate pages; separate further return number of properties URL

AndyHWChung:
…n number of properties URL


nubz (commented):
LGTM

PR #925, opened 2020-11-05
CPDP-2993: specs and removed userkey from helptext

nubz:


conf/messages
```diff
@@ -841,27 +841,24 @@ disposalMethod.personalRep.error.invalid=Select if the person sold it, gave it a
 
 disposalMethod.cyaChange=how the property was sold or disposed of
 
-wereYouAUKResident.main.title=Tell us about your residence status on the disposal date
```
AndyHWChung, line 844: we might as well delete these messages from the welsh file as well if they're not gonna be used anymore

nubz: will do

AndyHWChung:
LGTM

PR #983, opened 2021-03-10
DLS-3942-new Check that completion date is always on or after exchange date

rajanimohan22:
Check that completion date is always on or after exchange date

### DIFF
--- a/app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/triage/MultipleDisposalsTriageController.scala
+++ b/app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/triage/MultipleDisposalsTriageController.scala
@@ -854,7 +854,12 @@ class MultipleDisposalsTriageController @Inject() (
           viewConfig.enableFutureDateForDisposalAndCompletion,
           viewConfig.maxYearForDisposalsAndCompletion
         )
-        completionDateForm(maxDateAllowed)
+
+        val taxYearExchangedSelected: Option[TaxYearExchanged] =
+          answers.fold(_.taxYearExchanged, c => Some(c.taxYearExchanged))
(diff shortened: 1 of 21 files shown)


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/triage/MultipleDisposalsTriageController.scala
```diff
@@ -953,6 +958,13 @@ class MultipleDisposalsTriageController @Inject() (
       }
     }
 
+  private def getTaxYearByTaxYearExchanged(taxYearExhanged: Option[TaxYearExchanged]): Int =
+    taxYearExhanged match {
+      case Some(TaxYearExchanged.TaxYear2020) => 2020
+      case Some(TaxYearExchanged.TaxYear2021) => 2021
+      case _                                  => 0
```
calinursaciuc, line 965: Should this be None instead of zero?

calinursaciuc: Or is used somewhere as an Int?

conf/application.conf
```diff
@@ -261,7 +261,7 @@ enable-language-switching = true
 
 futureDate {
     disposalsAndCompletion {
-        enabled = true
+        enabled = false
```
calinursaciuc, line 264: Does Johnny needs this as true for time being for testing purposes?

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/triage/MultipleDisposalsTriageController.scala
rajanimohan22: Yeah Good idea, I will wrap the written value with Option.

PR #1002, opened 2021-03-25
DLS-4051: Backward compatibility for new SA question

rambabu-posa:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/TaxYear.scala
```diff
@@ -58,13 +58,12 @@ object TaxYear {
 
   implicit class TaxYearOps(private val taxTear: TaxYear) extends AnyVal {
 
-    def isItInLatestTaxYear(): Boolean = {
-      //TODO: remove below line and uncomment next line
-      val today = TimeUtils.today().plusYears(1L)
-      // val today = TimeUtils.today()
+    def isItInLatestTaxYear(futureTaxYearsEnabled: Boolean): Boolean = {
```
calinursaciuc, line 61: do we need a default of false?

rambabu-posa: As we enable or disable `futureDates` flag in `application.conf`, I think we will get both boolean values. If required, yes sure we can add default value

PR #1022, opened 2021-04-20
DCWL-118 Temporary Incorrect Calculation Error Message

rajanimohan22:
Temporary Incorrect Calculation Error Message

### DIFF
--- a/app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/account/home.scala.html
+++ b/app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/account/home.scala.html
@@ -34,7 +34,8 @@
   buttonLink: uk.gov.hmrc.cgtpropertydisposalsfrontend.views.html.components.button_link,
   sentReturnsList: uk.gov.hmrc.cgtpropertydisposalsfrontend.views.html.account.components.sent_returns_list,
   draftReturnsList: uk.gov.hmrc.cgtpropertydisposalsfrontend.views.html.account.components.draft_returns_list,
-  totalLeftHeading: uk.gov.hmrc.cgtpropertydisposalsfrontend.views.html.account.components.total_left_to_pay
+  totalLeftHeading: uk.gov.hmrc.cgtpropertydisposalsfrontend.views.html.account.components.total_left_to_pay,
+  warning: uk.gov.hmrc.cgtpropertydisposalsfrontend.views.html.components.warning
 )
 
 @(subscribed: Subscribed)(implicit request:RequestWithSessionData[_], messages: Messages, appConfig: ViewConfig)
@@ -123,7 +124,6 @@
   <div class="column-two-thirds account-header">
 
    @pageHeading(title, None)
-
    <h2 id="account-details">
     <span>
      @messages("account.home.accountName", subscribed.subscribedDetails.makeAccountName())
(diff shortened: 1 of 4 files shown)


conf/messages
```diff
@@ -215,6 +215,7 @@ accessibilityStatement.howWeTested.title=How we tested this service
 accessibilityStatement.howWeTested.p1=The service was tested on 15 May 2020 and checked for compliance with WCAG 2.1 AAA. It was built using parts that were tested by the <a href="{0}">Digital Accessibility Centre</a>.
 accessibilityStatement.howWeTested.p2=This page was prepared on 6 April 2020. It was last updated on 14 September 2020.
 
+warning.incorrectcalculation=Due to a temporary problem with the service, you cannot submit a return for the tax year 2021 to 2022 if one has already been sent for the tax year 2020 to 2021.If you submit a return, you may receive an incorrect calculation
```
rambabu-posa, line 218: Am expecting a space before "If" like
`2021. If`

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/account/home.scala.html
```diff
@@ -133,14 +133,18 @@ <h2 id="account-details">
     </span>
    </h2>
 
+   <p>
+    @warning(messages("warning.incorrectcalculation"))
+   </p>
```
rambabu-posa, line 138: This `<p></p>` tag is not required. Please check existing examples

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/landing_page.scala.html
```diff
@@ -123,8 +124,10 @@ <h2 class="heading-medium" id="beforeYouStart">@messages("landingPage.beforeYouS
         <li>@messages("landingPage.beforeYouStart.list1.bullet7")</li>
         <li>@messages("landingPage.beforeYouStart.list1.bullet8")</li>
         <li>@messages("landingPage.beforeYouStart.list1.bullet9")</li>
-    </ul>     
-    
+    </ul>
+    <p>
+        @warning(messages("warning.incorrectcalculation"))
+    </p>
```
rambabu-posa, line 130: I think, this <p></p> tag is not required. Please check existing examples. Probably please double check in UI

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/return_saved.scala.html
```diff
@@ -52,6 +52,10 @@
     @messages("draftReturnSaved.p2")
   </p>
 
+  <div>
+    @warning(messages("warning.incorrectcalculation"))
+  </div>
```
rambabu-posa, line 57: I think, this `<div></div>` tag is not required. Please check existing examples. Probably please double check in UI

conf/messages
rajanimohan22: Done. Never expected this would come from Content team. I have just copied.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/account/home.scala.html
rajanimohan22: Removed the <p> tag. I wanted to be inline with the existing tags.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/landing_page.scala.html
rajanimohan22: Removed the <p> tag. I wanted to be inline with the existing tags.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/return_saved.scala.html
rajanimohan22: This has to stay. Otherwise its causing unit tests to fail. There is one more warning message on the page similar to the new one we added. There is one unit test that selects 'doc.select("#content > article > div > strong")'. Both are matching and a unit test is failing. Instead of modifying the unit tests which will need to be re-modified once we remove this incorrectcalculation warning, I have adopted this approach.

conf/messages
calinursaciuc: Do we also need full stop at the end? Not important but just in case.

rambabu-posa: Yes I too agree need one full-stop at end.

rajanimohan22: Added the full stop to be inline with all other warning messages.

calinursaciuc (approved):

PR #1025, opened 2021-05-04
DCWL-126: Fix for Draft Return's liability issue.

rambabu-posa:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/services/returns/ReturnsService.scala
```diff
@@ -173,6 +181,94 @@ class ReturnsServiceImpl @Inject() (
                  else EitherT.rightT[Future, Error](())
     } yield unsentDraftReturnsWithTaxYearExchangedAndSAStatus
 
+  def updateCorrectTaxYearToSentReturns(
```
calinursaciuc, line 184: Shall this be names as "updateCorrectTaxYearToDraftReturns"?

rambabu-posa: They are sent returns only as we are referring `ReturnSummary`. We dont need to update "taxYear" field to draft returns.

calinursaciuc (approved):

PR #1052, opened 2021-08-13
DLS-4351: [RP] LR page-New drop down component with content and link

rambabu-posa:


rajanimohan22 (approved):
Changes reviewed and approved

PR #1062, opened 2021-09-22
DLS-4571: [RP] PRR dropdown box content updates.

rambabu-posa:


sureshhmrc (approved):
may be a unit spec helps to verify the content, but approving

PR #1089, opened 2022-01-18
DLS-4917: [RP] Draft Return's due date logic with Oct 27th, 2021.

rambabu-posa:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/config/ViewConfig.scala
```diff
@@ -185,4 +185,9 @@ class ViewConfig @Inject() (servicesConfig: ServicesConfig) extends Logging {
   val selfAssessmentTaxReturnsCorrections: String = getString("external-url.sa-tax-returns-corrections")
 
   val numberOfTaxYearsForReturns: Long = servicesConfig.getInt("returns-number-of-tax-years").toLong
+
+  val draftReturnNewDueDateStartYear: Int  = servicesConfig.getInt("draft-return.new-due-date.start-year")
```
lutnos, line 189: For myself, configuring a date, I would configure a String and parse that, rather than 3 integers

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/account/components/draft_list_item.scala.html
```diff
@@ -74,7 +75,17 @@
             })
 }
 
-@dueDate = @{completionDate._1.map(d => d.plusDays(30L))}
+@dueDate = @{
```
lutnos, line 78: Its nearly always better to put this kind of logic (which is calculating a value) outside of a view

lutnos (approved):

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/config/ViewConfig.scala
rambabu-posa: CGT has lots of configurations in this format.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/account/components/draft_list_item.scala.html
rambabu-posa: Yes I too agree with you. As its temporary solution and we may remove it later, I feel updating a view is a quick fix.

PR #1090, opened 2022-01-24
[DLS-4939]: local redirect fixed

ayub96:


lutnos (changes requested):
Note also the PR tests also fail

ayub96:
> Note also the PR tests also fail

I'll take a look

lutnos (approved):

PR #1105, opened 2022-05-19
[SM][DLS-5174] Moved dateFormatterForMultiDisposals logic to completionDateForm.

SeanMurray1:
…onDateForm.


rambabu-posa:
this TimeUtils class have two different utility functions

1. dateFormatter
2. dateFormatterForMultiDisposals

If we go through the code, we can understand that we are using this dateFormatterForMultiDisposals function only at only place, not for all multiple disposal journeys

that means we have written this function only to fix one scenario

As part of this ticket, please combine dateFormatterForMultiDisposals function logic into dateFormatter function. we should have only one function in TimeUtils for both single and multiple disposals.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/triage/MultipleDisposalsTriageController.scala
```diff
@@ -1939,17 +1939,126 @@ object MultipleDisposalsTriageController {
     Form(
       mapping(
```
rambabu-posa, line 1940: Please do not put this validation here.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/TimeUtils.scala
```diff
@@ -130,128 +130,6 @@ object TimeUtils {
 
     }
 
-  def dateFormatterForMultiDisposals(
```
rambabu-posa, line 133: integrate this `dateFormatterForMultiDisposals` function logic into `dateFormatter` and fix compilation issues and tests

PR #1115, opened 2022-07-19
[DLS-5467]- Updated Play 2.8.16 and bootstrap 5.24.0

JakeReid2020:


lutnos (changes requested):
I don't think this is right.  If you read the Confluence Blog post, it says that there is a vulnerability in the use of bindFromRequest, so they have changed its behaviour.

They say that in exceptional circumstances when the original behaviour is needed then use WithDefaultFormBinding.

You appear to have made a change to use it in all or most cases, so leaving us vulnerable.  Are all of our uses an exceptional case?  I believe this leaves us vulnerable

PR #1125, opened 2023-03-22
DLS-6922: new tax year rates

johno1985:
Added new message file entry and validation rules for 2023/2024 tax year


lutnos (approved):
Hmm, at some point we should be making this dynamic, but for now let us leave that as tech debt

johno1985:
Yes, I was thinking the same, and this frontend code is a bit of a can of worms

PR #1135, opened 2023-04-05
[DLS-7284] timedOut-page upgraded

ruthdas03:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/timed_out.scala.html
```diff
@@ -16,32 +16,21 @@
...
     @messages("timed-out.p1")
   </p>
 
-  <p>
-    @Html(messages("timed-out.p2"))
+  <p class="govuk-body">
+    @messages("timed-out.p2")
   </p>
 
-  @buttonLink(messages("timed-out.button"), routes.StartController.start().url)
```
nubz, line 45: it looks like this has not been replaced with anything - should be a GovukButton with an href value

nubz (approved):

PR #1195, opened 2023-06-12
[DLS-7549] Upgrade template annual_exempt_amount

ruthdas03:


test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/exemptionandlosses/ExemptionAndLossesControllerSpec.scala
```diff
@@ -1696,7 +1696,7 @@ class ExemptionAndLossesControllerSpec
               messageFromMessageKey(s"$key.title"),
               { doc =>
                 doc
-                  .select(s"#$key-form-hint")
+                  .select(s"#$key-govuk-hint")
```
nubz, line 1699: Hints that are generated by Govuk components use the format $key-hint, if these hints are from a Govuk component then remove the `govuk-` part of this - if these are hints from a custom component (such as unit_input) then we should be upgrading that to use GovukInput instead anyway

PR #1204, opened 2023-06-16
DLS-7598 CGT [returns/supportingevidence]: Upgrade CYA and Question

goodvibes:
https://jira.tools.tax.service.gov.uk/browse/DLS-7598

Updating to new layout

### DIFF
--- a/app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/supportingevidence/do_you_want_to_upload.scala.html
+++ b/app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/supportingevidence/do_you_want_to_upload.scala.html
@@ -48,6 +48,8 @@
 
 @hasErrors = @{form.hasErrors || form.hasGlobalErrors}
 
+@helpText = @{"helpText"}
+
 @layout(
   title,
   backLinkUrl = Some(backLink.url),
@@ -61,11 +63,11 @@
 
   <h1 class="govuk-heading-xl">@title</h1>
 
-  <p class="govuk-body-l">@messages(s"$key.help-text")</p>
+  <p class="govuk-hint" id=@helpText>@messages(s"$key.help-text")</p>
 
   @if(dueRepayment){
     @govukWarningText(WarningText(
-      iconFallbackText = Some("Warning"),
+      iconFallbackText = Some(messages("generic.warning")),
       content = Text(messages(s"$key.hint"))
     ))
   }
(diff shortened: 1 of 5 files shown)


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/supportingevidence/do_you_want_to_upload.scala.html
```diff
@@ -17,56 +17,93 @@
...
-                fieldId = key,
-                label = pageHeading(title),
-                errorKey = form.error(key).map(e => e.message),
-                hasErrors = hasErrors,
-                selected = form.value,
-                helpText = Some(Html(messages(s"$key.help-text"))),
-                yesContent = None,
-                warningField = if(dueRepayment){Some(messages(s"$key.hint"))}else{None}
-            )
+  <p class="govuk-body-l">@messages(s"$key.help-text")</p>
```
nubz, line 64: using a large font size makes the page appear differently than before - you should use `.govuk-hint` instead and also give the paragraph an id and use that in the fieldset's describedby value because currently this is now less accessible than before when the hint text was associated with the fieldset for screen readers - we can maintain that association by giving the hint text an id and using it in the describedby parameter passed to Fieldset()

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/supportingevidence/do_you_want_to_upload.scala.html
```diff
@@ -17,56 +17,93 @@
...
 
-            @submitButton(messages(if(isAmend) "button.continue" else "button.saveAndContinue"))
+  @if(dueRepayment){
+    @govukWarningText(WarningText(
+      iconFallbackText = Some("Warning"),
```
nubz, line 68: we should use `messages("generic.warning")` here as iconFallbackText because this is hard coding English for Welsh language users

nubz (changes requested):

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/supportingevidence/do_you_want_to_upload.scala.html
goodvibes: done

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/supportingevidence/do_you_want_to_upload.scala.html
goodvibes: done

PR #1220, opened 2023-07-05
[DLS-8283] cleanup-remove files not used

ruthdas03:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/views/returns/representee/is_first_return.scala.html
nubz, line 1: we should have kept this file - it was the upgraded file

nubz (approved):

PR #1244, opened 2023-10-11
DLS-8991 | Address app config warnings, move optimizely config to service level

mahesh-hmrc:
- Moved optimizely projectId to app config as it's applicable to all envs and makes sense for it be at service level
- Update `sbt-auto-build` plugin version

- Had an intermittent test failure and looks like when I opened that test in IDE to debug, picked up scalafmt change as part of that, and is unrelated to this story


TomRafferty:
got a failed PR build - intermittent failure?
changes look good, just need green pipeline :)

mahesh-hmrc:
Yeah, just an intermittent one. Successful now on rerun 👍

TomRafferty (approved):

PR #1246, opened 2023-10-31
DLS-8626 | Update rebase rule for 1982 & 2019

mahesh-hmrc:
Also addresses some of the compiler warnings
- Regarding `Xlint:-byname-implicit`, we are safe to do so considering the block it's applied to is a single expression


karolina-zaborska (approved):

LAJW (approved):
The code changes look sensible.

PR #1263, opened 2023-11-30
DLS-9370 | Handle acquisition method according to representative user journey

mahesh-hmrc:
For the journeys mentioned in https://docs.google.com/spreadsheets/d/15BS3r3ceZrhe0CYUi1tNJls8S3G8OWciuh3S64fl6EA/edit?pli=1#gid=1748147366, representative (capacitor/personalrepresentative) journey should bucket acquisition methods that isn't a `Bought`. 

Verified manually by running the journey. It's a risk to just cater for `Inherited` as highlighted in the ticket. Because, by choosing `Other` also fails. This change caters for both, and is inline with how we handle these acquisition methods in other parts of the same journey.

_Also includes changes for intermittent test failures. Involves fix to the test fixtures for some testsuites to factor in valid/invalid scenarios when generators are in play. Might not the be end of it all, but gets us a start and hopefully fix anything that comes along with next PRs._

### DIFF
--- a/test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/accounts/homepage/HomePageControllerSpec.scala
+++ b/test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/accounts/homepage/HomePageControllerSpec.scala
@@ -1988,6 +1988,7 @@ class HomePageControllerSpec
           case _: JustSubmittedReturn    => true
           case _: ViewingReturn          => true
           case _: FillingOutReturn       => true
+          case _: SubmittingReturn       => true
           case _                         => false
         }
       )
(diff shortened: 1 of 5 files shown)


test/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/accounts/homepage/HomePageControllerSpec.scala
```diff
@@ -1988,6 +1988,7 @@ class HomePageControllerSpec
           case _: JustSubmittedReturn    => true
           case _: ViewingReturn          => true
           case _: FillingOutReturn       => true
+          case _: SubmittingReturn       => true
```
johno1985, line 1991: Apologies for missing this

mahesh-hmrc: No problem 👍

johno1985 (approved):

PR #1267, opened 2023-12-08
DLS-9283: Update the content on the Property Acquisition Value page on the Multiple Disposal Journey

johno1985:


mahesh-hmrc (approved):
LGTM! 

Haven't had time to run through the journey and there is a bit of time pressure on this to make to QA. So, have discussed with Matt and that I could approve it from code point of view and let QA pickup for journey verification.

As have been keen to learn the journey for any of the changes, I'll continue to do that in parallel to any QA work being done. This is for own learnings being new to the user journeys. Thanks

PR #1272, opened 2023-12-27
DLS-9243 - updated links by adding govuk-link class

Timidigital:
[DLS-9243](https://jira.tools.tax.service.gov.uk/browse/DLS-9243)


ruthdas03 (changes requested):
Would need a similar change in the message.cy

ruthdas03 (approved):

PR #1274, opened 2024-01-08
[DLS-9271] Address sorted and fixed test

ruthdas03:


test/uk/gov/hmrc/cgtpropertydisposalsfrontend/services/UKAddressLookupServiceImplSpec.scala
```diff
@@ -297,21 +297,21 @@ class UKAddressLookupServiceImplSpec extends AnyWordSpec with Matchers with Mock
             List(
               UkAddress(
                 "line1",
-                Some("line2"),
```
nubz, line 300: did this spec need to change at all? why are the addresses used here being updated?

ruthdas03: So effectively, the addresses in the list are not updated but arranged in sorted order.

nubz (approved):

PR #1291, opened 2024-06-04
[DLS-10022] Added logs for 400 errors

ruthdas03:


TomRafferty (approved):

TomRafferty (commented):
the sbt changes can just be removed as it seems they are added then removed anyway so it'll just be clutter in the git history - other than that, all looks good 😄

ruthdas03:
> the sbt changes can just be removed as it seems they are added then removed anyway so it'll just be clutter in the git history - other than that, all looks good 😄

Done

TomRafferty (approved):
nice one

PR #1292, opened 2024-06-10
DLS-10424 Partial removal of shapeless and saving shapeless results to speed up compilation

LAJW:
Focusing on the following step: "Test / clean" "Test / compile"

Before: 290-300s
After: 85s

Memory consumption stays under 700mb.

We can go further, but that's a start.


johno1985:
rebuild

LAJW:
Superseded by: https://github.com/hmrc/cgt-property-disposals-frontend/pull/1293

PR #1304, opened 2024-08-09
[DLS-10782] Fix multiple form submission

ruthdas03:


johno1985 (changes requested):
This is incorrect. The ticket is to prevent multiple submission, not double click, which is already done by the platform

PR #1307, opened 2024-08-23
DLS-10792 Parse JSON in Connectors and Wiremock Tests

LAJW:


gajalakshmiseetharaman (approved):
I have reviewed the changes which are working in my local syatem. Hence approving.

PR #1308, opened 2024-08-30
[DLS-10782] Fix multiple form submission

ruthdas03:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/onboarding/SubscriptionController.scala
```diff
@@ -69,6 +69,27 @@ class SubscriptionController @Inject() (
...
+                                            addressSource = details.addressSource,
+                                            contactNameSource = details.contactNameSource
+                                          ),
+                                          request.subscriptionReady.ggCredId
+                                        )
+                                      )
+                                    )
+                                  )
+                                )
         subscriptionResponse <-
```
LAJW, line 93: 🚫 This should be surrounded by a lock and there should be a check for the registration status.

LAJW (changes requested):

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/onboarding/SubscriptionController.scala
ruthdas03: Not sure why we want to have lock on a subscribe journey twice. We are already using lock on a subscribe call at the backend.

LAJW: Rephrased.

ruthdas03:
Updated journey status for Registration flow. 
However not sure about adding 'lockservice'  in the Registration Controller.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/JourneyStatus.scala
```diff
@@ -253,6 +253,15 @@ object JourneyStatus {
       )
     }
 
+    final case class Registering(
+      registrationDetails: RegistrationDetails,
+      ggCredId: GGCredId
+    ) extends RegistrationStatus {
+      val emailSource: Option[EmailSource] = Some(
+        registrationDetails.emailSource
```
johno1985, line 261: I'm guessing that Registering only needs to extend JourneyStatus

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/StartController.scala
```diff
@@ -174,7 +175,7 @@ class StartController @Inject() (
     request: RequestWithSessionDataAndRetrievedData[AnyContent]
   ): Future[Result] =
     journeyStatus match {
-      case _: Subscribed =>
+      case _: Subscribed | _: Registering =>
```
johno1985, line 178: the lock should go on the repository layer, where `journeyStatus` comes from

LAJW: Not necessarily, because you want to send between status updates. IMO it should be pretty high.

johno1985: on second thought, i'm wondering if we need the lock at all, since updates are already done atomically with `findOneAndUpdate`

LAJW: We don't need locks on individual database operations - find and update is already atomic.

What you need a lock for is to create a transaction. Transactions are usually done in the application layer, not the database layer.

We want to implement the following bit of pseudocode, in the snippet above:

```scala
withLock {
  status <- getJourneyStatusFromCache
  if (status.type == SubmissionInProgress) {
    _ <- sendCreateAccountRequest
    _ <- sendEnrolmentRequest
    _ <- setJourneyStatusToComplete
  }
} match {
  case None => retryLock
  case Some(_) => Success
}
```

This ensures only one account creation request is done at any given time. Without it, the double submission, when performed quickly enough would still occur, as Ruth's noticed (two or more ifs can be true before you set the status journey to complete).

LAJW: I'm still not fully happy with it - if there's a crash when this chain of futures executes, the application will be left in an inconsistent state and rerunning the request may still create two accounts. But it's still miles ahead of what we have right now.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/JourneyStatus.scala
johno1985: So it probably does not need the extra attributes

johno1985 (commented):
Provisionally approved pending discussion with Mahesh

johno1985 (approved):

johno1985 (approved):

PR #1319, opened 2024-11-27
DLS-4071: Email regex change for validations that comply with ui AT t…

vamsikrishnapujari:
…ests.

### DIFF
--- a/app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/email/Email.scala
+++ b/app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/email/Email.scala
@@ -31,7 +31,7 @@ final case class Email(value: String) extends AnyVal
 object Email {
 
   val emailRegex: Predicate[String] =
-    "^(?=[^\\s]{3,132}$)[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\\.[a-zA-Z]{2,}$".r.pattern.asPredicate()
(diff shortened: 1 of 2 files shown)


johno1985:
Please address pr bot comment

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/email/Email.scala
```diff
@@ -30,7 +30,8 @@ final case class Email(value: String) extends AnyVal
 
 object Email {
 
-  val emailRegex: Predicate[String] = "^(?=.{3,132}$)[^@]+@[^@]+$".r.pattern.asPredicate()
+  val emailRegex: Predicate[String] =
```
johno1985, line 33: shouldn't we be using the play built in email validator?

mahesh-hmrc: Only caveat with the underlying constraint is that the regex isn't accessible. And, for this particular usecase we have a service layer that requires access to this regex. Though the logic in service layer seems to be a workaround to an issue that might have happened, am reluctant to drop that.

The only way to get this going is to then inline the regex from the play data validator email constraint to our repo. I would avoid that as it doesn't give us anything different to where we are now.

And, we also would like to use same regex when generating random email in ui-tests. We currently require a regex there aswell. So, will be inlining in multiple places.

Also, major issue seems to be that the play email validator doesn't validate for missing domain part of email. @vamsikrishnapujari found that in his testing.

So, I suggest we use something that works for our usecase here and a pattern we have tested thoroughly

ruthdas03 (approved):

PR #1323, opened 2025-01-02
DLS-11140 : Supporting document upload file validation

gajalakshmiseetharaman:
DLS-11140 : Supporting document upload file validation


conf/messages
```diff
@@ -130,6 +130,7 @@ generic.no=No
 generic.success=Success
 generic.errorPrefix=Error:
 generic.error=Error
+generic.errorSummaryHeader=There is a problem
```
ruthdas03, line 133: We need Welsh for this error message, it's missing.

ruthdas03 (changes requested):

conf/messages
gajalakshmiseetharaman: Thanks Ruth, Welsh translation added.

ruthdas03 (approved):

PR #1365, opened 2026-08-04
DLS-13320 Duplicate file issue

ruthdas03:


app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/supportingevidence/SupportingEvidenceController.scala
```diff
@@ -408,8 +408,22 @@ class SupportingEvidenceController @Inject() (
     hc: HeaderCarrier
   ): EitherT[Future, Error, StoreUpscanSuccessResult] = {
 
+    def mandatoryEvidence: Option[MandatoryEvidence] =
```
mahesh-hmrc, line 411: Though is a correct change, but, it only assumes a specific direction for validation. i.e., that a mandatory evidence always exists by the time user lands at supporting evidence. Which is not always the case, so, this validation should also happen on mandatory evidence side. That means, we need to find a better place for these validations to avoid duplicating checks. Maybe `DraftReturn` is better place for these.

Also, `.get` is not safe here

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/yeartodatelliability/YearToDateLiabilityController.scala
```diff
@@ -1986,15 +1987,32 @@ class YearToDateLiabilityController @Inject() (
...
-            },
-            upscanUpload => handleUpscanCallback(upscanUpload.upscanCallBack)
-          )
+
+            case Right((newUpscanUpload, storeResult)) =>
+              storeResult match {
+                case DuplicateUpscanFileName =>
+                  BadRequest(
+                    mandatoryEvidencePage(
+                      newUpscanUpload,
```
mahesh-hmrc: We can reuse this upload object here as it contains the  successful callback from upscan. This should initiate a new upscan call and then use the upload object returned. Please see how supporting evidence manages it [here](https://github.com/hmrc/cgt-property-disposals-frontend/pull/1365/changes#diff-7b4979e9ea8c57ea4af90d570a1af45a5b83df63bd6c1c3be09502eada9d00a7R365).

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/supportingevidence/SupportingEvidenceController.scala
```diff
@@ -409,7 +424,9 @@ class SupportingEvidenceController @Inject() (
   ): EitherT[Future, Error, StoreUpscanSuccessResult] = {
 
     val duplicateFileName =
```
mahesh-hmrc: It would make more sense to have this check on `DraftReturn` so that this verification can be done in both mandatory and supporting evidence. That way, current check in mandatory evidence controller can be simplified.

Also, there is a gap in logic still for draft returns. We check for duplicates in supporting and mandatory evidence controllers. But, if a user had these duplicates in draft return already before this fix, submission will still fail. So, we need to also check for this duplicate in `CheckAllAnswersAndSubmitController`.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/returns/DraftReturn.scala
```diff
@@ -250,6 +253,26 @@ object DraftReturn {
 
   }
 
+  def supportingEvidence(draftReturn: DraftReturn) =
+    draftReturn.supportingEvidenceAnswers.fold(None: Option[SupportingEvidence]) {
```
mahesh-hmrc: This should be an empty list returned than `None` which isn't the correct type for it.

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/controllers/returns/yeartodatelliability/YearToDateLiabilityController.scala
```diff
@@ -2463,14 +2481,24 @@ class YearToDateLiabilityController @Inject() (
...
-    for {
-      _ <- returnsService.storeDraftReturn(newJourney)
-      _ <- EitherT(
-             updateSession(sessionStore, request.toSession)(
-               _.copy(journeyStatus = Some(newJourney))
+
+    val mandatoryEvidenceFile: String       = mandatoryEvidence(newDraftReturn).map(_.fileName).getOrElse("")
+    val supportingEvidenceFileList: List[String] = supportingEvidence(newDraftReturn).iterator.map(_.fileName).toList
+
+    if (supportingEvidenceFileList.contains(mandatoryEvidenceFile)) {
```
mahesh-hmrc: Inline with support evidence controller comment, pls use the check from DraftReturn once duplicate logic is moved there.

vamsikrishnapujari (approved):

app/uk/gov/hmrc/cgtpropertydisposalsfrontend/models/returns/DraftReturn.scala
```diff
@@ -250,6 +253,33 @@ object DraftReturn {
...
 
+  private def supportingEvidence(draftReturn: DraftReturn): List[SupportingEvidence] =
+    draftReturn.supportingEvidenceAnswers.fold(List.empty) {
+      case completeSupportingEvidenceAnswers: CompleteSupportingEvidenceAnswers     =>
+        completeSupportingEvidenceAnswers.evidences
+      case incompleteSupportingEvidenceAnswers: IncompleteSupportingEvidenceAnswers =>
+        incompleteSupportingEvidenceAnswers.evidences
+    }
+
+  def mandatoryEvidence(draftReturn: DraftReturn): Option[MandatoryEvidence] =
```
ShivangiShah42: This can be private as not being used outside of DraftReturn

ShivangiShah42 (approved):

